PR #38, opened 2017-02-24
Use batch_states for ACER

kiyukuta:
I modified as other agents.


kiyukuta:
Sorry, I just noticed continuous ACER is WIP.
I will modify and resend this PR after the continuous one is merged !

PR #71, opened 2017-04-21
Fix usage of a term "frequency"

muupan:
Actually, It is used to mean "interval" for now, so this PR will replace it with "interval".

This PR might affect existing code because names of some arguments will be changed.


muupan:
All the tests passed.

PR #81, opened 2017-04-26
REINFORCE agent

### LINKED ISSUE #70: REINFORCE
A simple REINFORCE implementation that doesn't require a value function would be helpful.

muupan:
Resolves #70


muupan:
all the tests passed

PR #121, opened 2017-07-27
Remove ALE dependency

muupan:
Since atari_py has been updated to use the latest ALE, we should no longer need to install the ALE separately. I think it is better to confirm there is no performance regression, though.


muupan:
I ran `python examples/ale/train_dqn_ale.py breakout` for current master and this branch. It seems there's no clear performance regression.

scores txt

chainerrl/envs/ale.py
```diff
@@ -57,10 +65,10 @@ def __init__(self, game, seed=None, use_sdl=False, n_last_screens=4,
         # atari_py is used only to provide rom files. atari_py has its own
         # ale_python_interface, but it is obsolete.
         if not atari_py_available:
-            raise RuntimeError('You need to install atari_py to use ALE.')
+            raise RuntimeError('You need to install atari_py>=0.1.1 to use ALE.')  # NOQA
```
muupan: Fixed, thank you.

PR #124, opened 2017-07-31
Support environments without git command

### LINKED ISSUE #123: In case git is not installed, catch exception in `is_return_code_zero` method
When git command is not installed, `FileNotFoundError` will be thrown on def `is_under_git_control()` method.  I think the exception should be catched in `is_return_code_zer` method.

https://github.com/chainer/chainerrl/blob/d56e54dae3de908078e33b44c931be6a516af85e/chainerrl/experiments/prepare_output_dir.py#L17-L32

This error can occur often when we use docker.

muupan:
Resolves #123


toslunar:
LGTM

PR #128, opened 2017-08-03
support python3.4 for random.sample with deque

knorth55:
with `Python3.4`, it causes error below.
```python
  File "quickstart.py", line 111, in <module>
    action = agent.act_and_train(obs, reward)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/agents/dqn.py", line 340, in act_and_train
    self.replay_updator.update_if_necessary(self.t)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/replay_buffer.py", line 194, in update_if_necessary
    transitions = self.replay_buffer.sample(self.batchsize)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/replay_buffer.py", line 42, in sample
    return random.sample(self.memory, n)
  File "/opt/rl/lib/python3.4/random.py", line 311, in sample
    raise TypeError("Population must be a sequence or set.  For dicts, use list(d).")
TypeError: Population must be a sequence or set.  For dicts, use list(d).
```

I know `Python3.4` is not officially supported, but I only have this bug with `python3.4` and everything works fine, so i thought it is pity to remain this and block `python3.4` users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,12 @@ language: python
 cache: pip
 python:
   - "2.7"
+  - "3.4"
   - "3.5.1"
+env:
+matrix:
+  allow_failures:
+    - python: "3.4"
 # command to install dependencies
 install:
   - pip install --upgrade pip setuptools wheel
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -6,6 +6,8 @@ from builtins import *  # NOQA
 from future import standard_library
 standard_library.install_aliases()
 from collections import deque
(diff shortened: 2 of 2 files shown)


muupan:
I think it is not good to introduce unnecessary copy for python2 and >=3.5, since replay buffer can be large. Can you check the python version and wrap it with list only when needed?

knorth55:
Ok I updated to check python version.

knorth55:
I add `python3.4` as `allow_failures` for `coverage`.
Was it good?

PR #130, opened 2017-08-08
Fix save/load in EpisodicReplayBuffer

### LINKED ISSUE #129: Save/Load in EpisodicReplayBuffer
As for EpisodicReplayBuffer, not only self.memory but also self.epsodic_memory should be saved/loaded I think.

toslunar:
Resolves #129


muupan:
LGTM

PR #131, opened 2017-08-09
Use faster queue for replay buffers

### LINKED ISSUE #36: The tutorial code causes TypeError on python 3.4
On python 3.4, random.sample don't accept collections.deque, so I got such error.

```py3
  File "quickstart.py", line 111, in <module>
    action = agent.act_and_train(obs, reward)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/agents/dqn.py", line 340, in act_and_train
    self.replay_updator.update_if_necessary(self.t)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/replay_buffer.py", line 194, in update_if_necessary
    transitions = self.replay_buffer.sample(self.batchsize)
  File "/opt/rl/lib/python3.4/site-packages/chainerrl/replay_buffer.py", line 42, in sample
    return random.sample(self.memory, n)
  File "/opt/rl/lib/python3.4/random.py", line 311, in sample
    raise TypeError("Population must be a sequence or set.  For dicts, use list(d).")
TypeError: Population must be a sequence or set.  For dicts, use list(d).
```

python 2.7 works fine, and meybe 3.5+.

toslunar:
TODO:
- [x] Add tests
- [x] Measure performance

This will happen to resolve #36 (see also #128)

### DIFF
--- a/chainerrl/misc/collections.py
+++ b/chainerrl/misc/collections.py
@@ -1,4 +1,5 @@
 import itertools
+
 import numpy as np
 
 
--- a/chainerrl/misc/collections.py
+++ b/chainerrl/misc/collections.py
@@ -1,7 +1,5 @@
 import itertools
 import numpy as np
-import random
-import six
 
 
 """
@@ -9,7 +7,10 @@ def _sample_n_k(n, k):
     return random.sample(six.moves.range(n), k)
 """
 
+
 def _sample_n_k(n, k):
+    if not 0 <= k <= n:
+        raise ValueError("Sample larger than population or is negative")
     if 3 * k >= n:
         return np.random.choice(n, k, replace=False)
     else:
@@ -63,7 +64,6 @@ def _sample_n_k(n, k):
             selected_add(x)
         return result[0]
         """
-
 
 
 class RandomAccessQueue(object):
@@ -162,13 +162,6 @@ class RandomAccessQueue(object):
         return self._queue_front.pop()
 
     def _sample(self, k):
-        n = len(self)
-        if k > n:
-            raise ValueError("Sample larger than population or is negative")
-
-        # The following code is equivalent to
-        # return [self[i] for i in np.random.choice(n, k, replace=False)]
-
         nf = len(self._queue_front)
         return [self._queue_front[i] if i < nf else self._queue_back[i - nf]
-                for i in _sample_n_k(n, k)]
+                for i in _sample_n_k(len(self), k)]
--- a/chainerrl/misc/collections.py
+++ b/chainerrl/misc/collections.py
@@ -1,5 +1,6 @@
 import itertools
-import numpy as np
+import random
+import six
 
 
 class RandomAccessQueue(object):
@@ -103,8 +104,8 @@ class RandomAccessQueue(object):
             raise ValueError("Sample larger than population or is negative")
 
         # The following code is equivalent to
-        # return [self[i] for i in np.random.choice(n, k)]
+        # return [self[i] for i in random.sample(six.moves.range(n), k)]
 
         nf = len(self._queue_front)
         return [self._queue_front[i] if i < nf else self._queue_back[i - nf]
-                for i in np.random.choice(n, k)]
+                for i in random.sample(six.moves.range(n), k)]
--- a/chainerrl/misc/collections.py
+++ b/chainerrl/misc/collections.py
@@ -1,3 +1,11 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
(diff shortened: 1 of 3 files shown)


toslunar:
😰
before: 13.168848936009454
after: 16.709212244983064

```python
import numpy as np
import timeit

from chainerrl import replay_buffer


def rand_state():
    return np.random.rand(1, 3, 50, 20).astype(np.float32)


def rand_action():
    return np.random.rand(1, 40).astype(np.float32)


def rand_reward():
    return np.random.rand()


def f(capacity, batch_size, steps, replay_start_size):
    rbuf = replay_buffer.ReplayBuffer(capacity=capacity)
    s = rand_state()
    a = rand_action()
    for i in range(steps):
        next_s = rand_state()
        next_a = rand_action()
        rbuf.append(s, a, rand_reward(), next_s, next_a)
        s = next_s
        a = next_a
        if i >= replay_start_size:
            rbuf.sample(batch_size)


print(min(timeit.Timer(
    'f(10000, 64, 100000, 1000)',
    setup="from __main__ import f; gc.enable()").
    repeat(repeat=3, number=1)))
```

toslunar:
before: 337.00381314000697
after: 171.62339709801017
for `f(100000, 64, 1000000, 10000)`.

muupan:
Good job! But this solution seems unecessarily complex to me. How about using a ring buffer like this? https://github.com/matthiasplappert/keras-rl/blob/master/rl/memory.py#L35

I think we can assume `maxlen` is specified for replay buffers, otherwise training would face out-of-memory eventually.

toslunar:
`EpisodicReplayBuffer` limits the number of the transitions in the buffer.  It's not easy to determine the best `maxlen` of `self.episodic_memory = RingBuffer(maxlen)`.  Of course, there is sufficiently large one (and this doesn't cost memory much, relatively to other parts): `self.episodic_memory = RingBuffer(capacity)`.

toslunar:
- collections.deque + random.sample: 8.064196354011074
```python
q = deque(maxlen=10000)
for _ in range(100000):
    q.append(1)
    if len(q) > 1000:
        random.sample(q, 64)
```

- RandomAccessQueue
-- at d7299fcec49965b65426e4d36c11870c8af2f9a0: 13.529830269049853
-- at fc39b2098d37d03f16bf340e93ddf823dc20aa02: 10.133344302070327
-- at b8bec91de30e3f6f9f9e84793cdd48006be18421: 5.198489462025464 
```python
q = RandomAccessQueue(maxlen=10000)
for _ in range(100000):
    q.append(1)
    if len(q) > 1000:
        q._sample(64)
```

toslunar:
time (secs) after the debug:

|(steps, maxlen)|baseline: collections.deque + random.sample|collections.deque + _sample_n_k|RandomAccessQueue + _sample_n_k|
|-|-|-|-|
|(100000, 10000)|7.3263045670464635|5.774126871023327|7.128595568938181|
|(300000, 30000)|26.707710441900417|21.629545310977846|20.986778931925073|
|(1000000, 100000)||197.8464105380699|74.75397148309276|

Non-repetitive sampling by `[q[i] for i in _sample_n_k(len(q),64)]` is faster than by `random.sample(q, 64)`.  With this faster sampling, `RandomAccessQueue` is faster than `collections.deque` when `maxlen` is larger than around 30000.

tests/misc_tests/test_collections.py
```diff
@@ -0,0 +1,212 @@
...
+        n = len(xs)
+
+        sorted_xs = np.sort(xs)
+        cdfs = (1 + special.erf((sorted_xs - mean)
+                                / (np.sqrt(2) * std))) / 2
+        cdfs_l = (np.arange(n)) / n
+        cdfs_r = (np.arange(n) + 1) / n
+
+        # Kolmogorov-Smirnov statistic
+        d = max(np.amax(np.abs(cdfs - cdfs_x)) for cdfs_x in [cdfs_l, cdfs_r])
```
muupan: scipy has `scipy.stats.kstest`, so I think it's better to use it.

You can get p-value by `scipy.stats.kstest(xs, 'norm', args=(mean, std))[1]`.

chainerrl/misc/collections.py
```diff
@@ -0,0 +1,129 @@
+import itertools
```
muupan: Add future imports (at least `range` is affected)

toslunar:
Fixed.

chainerrl/misc/collections.py
okuta: Fixed.

PR #133, opened 2017-08-12
Fix REINFORCE's missing initialization of t

muupan:
REINFORCE doesn't have code to initialize `self.t`. The problem has been not found because `chainer.experiemnts.train_agent` sets it, but agents should be self-contained.


toslunar:
Nice catch and good job.

BTW, ChainerRL doesn't have rules for `Agent.t` (see e.g. #135), while I believe the PR does the best at this moment.

PR #137, opened 2017-08-16
[WIP] Improve batch_states to support cupy.ndarray as observations

muupan:
This PR enables `batch_states` to accept `cupy.ndarray`s so that you can use `cupy.ndarray` as observations.


muupan:
This coverage decrease is due to gpu-only code, so I think it's OK.

PR #140, opened 2017-08-19
[policies/gaussian_policy] Improve docstring

iory:
Deleted var_type from FCGaussianPolicyWithFixedCovariance 's docstring, 
because that's argument does not exist.
https://github.com/chainer/chainerrl/blob/master/chainerrl/policies/gaussian_policy.py#L130-L154


muupan:
Thank you for the fix!

PR #141, opened 2017-08-29
Use F.matmul instead of F.batch_matmul

toslunar:
batch_matmul is going to be unavailable in Chainer v3, or would be deprecated even if https://github.com/chainer/chainer/pull/3016 is merged.

### DIFF
--- a/chainerrl/misc/chainer_compat.py
+++ b/chainerrl/misc/chainer_compat.py
@@ -1,12 +1,15 @@
-from distutils.version import StrictVersion
+try:
+    from packaging import version
+except ImportError:
+    from pip._vendor.packaging import version
 
 import chainer
 import chainer.functions as F
 
 
-chainer_version = StrictVersion(chainer.__version__)
+chainer_version = version.parse(chainer.__version__)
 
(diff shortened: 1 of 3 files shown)


toslunar:
Oops, the PR depends on the new F.matmul in v3.

muupan:
I think It is better to support both v2 and v3 for now by checking the version of chainer and use F.numpy_like_matmul for v2.

muupan:
Can you merge the current master to this PR so that CI runs with both Chainer v2 and v3?

chainerrl/misc/chainer_compat.py
```diff
@@ -0,0 +1,20 @@
...
+import pkg_resources
+
+import chainer.functions as F
+
+
+chainer_version = StrictVersion(
+    pkg_resources.get_distribution("chainer").version)
+
+if chainer_version < StrictVersion('3.0.0a1'):
+    def matmul_v3(a, b, **kwargs):
```
muupan: Can you add a docstring to explain why this is needed?

toslunar:
Rebased to the current master

muupan:
LGTM

PR #144, opened 2017-09-20
Enhance plotting tool

toslunar:
The PR enables to plot intervals as filled regions, supporting the formats
- `--y-range err` (e.g. [mean - stdev, mean + stdev])
- `--y-range low,high` (e.g. [min, max])
tmp

It also add arguments specifying
- x-axis
- name of saved png (or None to show it on the display)

### DIFF
--- a/tools/plot_scores.py
+++ b/tools/plot_scores.py
@@ -8,13 +8,13 @@ import argparse
 import os
 
 import matplotlib
-matplotlib.use('Agg')  # Needed to run without X-server
-import matplotlib.pyplot as plt
+# Don't `import matplotlib.pyplot` here
 import pandas as pd
 
 
 def main():
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


tools/plot_scores.py
```diff
@@ -8,41 +14,77 @@
...
     parser.add_argument('--label', action='append', dest='labels',
                         default=[], type=str,
                         help='specify labels for scores.txt files')
+    parser.add_argument('--x', type=str, default='steps')
+    parser.add_argument('--y-range', type=str, default=None)
+    parser.add_argument('--fill-alpha', type=float, default=0.3)
     args = parser.parse_args()
 
+    if args.savefile:
+        matplotlib.use('Agg')  # Needed to run without X-server
```
okuta: How about using environment variable `MPLBACKEND=Agg` ?
Please see chainer/chainercv#355

PR #147, opened 2017-10-09
Use DQN.act method in DQN.act_and_train to remove code duplication

delta2323:
The first part of `DQN.act_and_train` is exactly identical to what `DQN.act` does. This PR changes `DQN.act_and_train` to use its `act` method to get a greedy action.


toslunar:
To remove the code duplication, it seems good to define a method that returns `action_value`, which would be called from `act` and `act_and_train`.

delta2323:
I missed that `act_and_train` uses `action_value`.

PR #148, opened 2017-10-09
Use Chain.init_scope in the quick start

delta2323:
Initialization of links with `Chain.__init__` is obsolete from Chainer v2. This PR changes to use `Chain.init_scope` instead in the quick start. As some I ran the code to confirm it works well, some results have changed (mainly caused by the change of random seed). Let me know if it is troublesome.


toslunar:
LGTM

PR #155, opened 2017-10-17
Fix episodic buffer __len__

### LINKED ISSUE #138: Performance of ACER compared to A3C
(Opening a separate issue since it's not relevant to #125.)
The performance of ACER isn't what it should be, compared to A3C. I've run them both on a variety of games, and while ACER doesn't perform poorly per se, it doesn't improve on the sample efficiency of A3C.*

I cloned a fresh copy of chainerrl to be sure I hadn't changed anything, then downloaded the PyPI version in case there was a recent change causing problems, but I didn't see any noticeable change in either case.

I'll search through agents/acer.py for bugs just in case, but there might easily be something wrong on my end. Any guess whether the underlying system might affect performance to this degree?

I've been using a Debian 8.8 machine with two 8-core processors. I used 16 worker threads and didn't use an LSTM layer. My Python libraries are installed in a Conda environment. I'd expect more catastrophic failures from incorrectly installed dependencies, so I don't think that could be the problem.

*(Which is important to me because I need an efficient off-policy algorithm.)

muupan:
- Add `AbstractReplayBuffer` and `AbstractEpisodicReplayBuffer` to clarify the interfaces of replay buffers
- Change the behaviour of the `__len__` of `EpisodicReplayBuffer` and `PrioritizedEpisodicReplayBuffer` so that they now return the number of transitions, not of episodes.
- Add the `n_episodes` property to count the number of episodes
- Update the examples accordingly
- Update the tests accordingly (and fix a bug in tests)

Resolves #138

### DIFF
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -407,6 +407,11 @@ class ReplayUpdater(object):
     def update_if_necessary(self, iteration):
         if len(self.replay_buffer) < self.replay_start_size:
             return
+
+        if (self.episodic_update
+                and self.replay_buffer.n_episodes < self.batchsize):
+            return
+
         if iteration % self.update_interval != 0:
             return
 
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -115,17 +115,16 @@ class AbstractEpisodicReplayBuffer(AbstractReplayBuffer):
     def stop_current_episode(self):
         """Notify the buffer that the current episode is interrupted.
 
-        When a transtion with is_state_terminal=True is appended, the buffer
-        will treat it as the last transition in an episode and the next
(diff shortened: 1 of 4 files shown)


chainerrl/replay_buffer.py
```diff
@@ -14,14 +14,16 @@
 from chainerrl.misc.prioritized import PrioritizedBuffer
 
 
-class ReplayBuffer(object):
+class AbstractReplayBuffer(object):
```
toslunar: Why don't you decorate the methods with `@abstractmethod`?

muupan:
Thank you for your review! I added a check of `n_episodes` in `ReplayUpdater.update_if_necessary` as well and followed your suggestions.

toslunar:
LGTM!

Issue #132, opened 2017-08-12
Better default of phi


muupan:
The default value`phi`, a feature extractor applied to an observation, is an identity. Since most agents assume numpy.float32, almost always you need to specify the converter as phi when you use gym-like envs. I think it is helpful to apply this conversion as default.

Issue #79, opened 2017-04-24
Extend gym.Wrapper instead of env_modifiers


muupan:
Since gym has introduced its own interface to modify envs using gym.Wrapper, I think it is better to use it in ChainerRL instead of directly modifying methods as in env_modifiers.

toslunar:
It seems `gym.Wrapper` assumes the base environment is a subclass of `gym.Env`.  For example,
```
gym.wrappers.TimeLimit(chainerrl.envs.abc.ABC(), max_episode_steps=2)
```
raises `AttributeError: 'ABC' object has no attribute 'metadata'`.

PR #160, opened 2017-11-07
Add wrappers to cast observations

muupan:
This PR adds two env wappers that inherit gym.Wrapper: `CastObservation`, `CastObservationToFloat32`.

I think using these wrappers instead of passing `phi` makes code easier to read, so the examples are also updated accordingly.

Related to https://github.com/chainer/chainerrl/issues/79 and https://github.com/chainer/chainerrl/issues/132


chainerrl/wrappers/cast_observation.py
```diff
@@ -0,0 +1,38 @@
...
+        env: Env to wrap.
+        dtype: Data type object.
+    """
+
+    def __init__(self, env, dtype):
+        super().__init__(env)
+        self.dtype = dtype
+
+    def _observation(self, observation):
+        self.original_observation = observation
```
toslunar: Could you document the attribute, if it'll be provided as a public one?

muupan:
Thank you for the review! I resolved all the points.

PR #162, opened 2017-11-07
Return dict for info of env.step

### LINKED ISSUE #161: ABC's info returned by env.step is None
https://github.com/chainer/chainerrl/blob/master/chainerrl/envs/abc.py#L96

It should be `dict` instead according to the docstring of gym.Env.

https://github.com/openai/gym/blob/master/gym/core.py#L94

toslunar:
Resolves #161


muupan:
LGTM

PR #166, opened 2017-11-09
Add a utility function to draw a computational graph

muupan:
This PR adds `chainer.misc.draw_computational_graph` to help draw computational graphs of RL models as PNG images.

I'm not sure if I should draw a graph in every example, but it would be a good habit. For this PR I just add such code only in `examples/gym/train_reinforce_gym.py`

### DIFF
--- a/chainerrl/misc/draw_computational_graph.py
+++ b/chainerrl/misc/draw_computational_graph.py
@@ -42,8 +42,8 @@ def draw_computational_graph(outputs, filepath):
     """Draw a computational graph and write to a given file.
 
     Args:
-        outputs (object): Output(s) of the computational graph. Each
-            item must be Variable, ActionValue, Distribution or list of them.
+        outputs (object): Output(s) of the computational graph. It must be
+            a Variable, an ActionValue, a Distribution or a list of them.
         filepath (str): Filepath to write a graph without file extention.
             A DOT file will be saved with ".gv" extension added.
             If Graphviz's dot command is available, a PNG file will also be
--- a/chainerrl/misc/draw_computational_graph.py
+++ b/chainerrl/misc/draw_computational_graph.py
@@ -1,0 +1,54 @@
+from __future__ import division
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import absolute_import
+from builtins import *  # NOQA
+from future import standard_library
+standard_library.install_aliases()
+
+import subprocess
+import tempfile
+
+import chainer.computational_graph
+import chainerrl
+
+
+def collect_variables(obj):
+    variables = []
+    if isinstance(obj, chainer.Variable):
+        return [obj]
+    elif isinstance(obj, chainerrl.action_value.ActionValue):
+        return [obj.greedy_actions(), obj.max()]
+    elif isinstance(obj, chainerrl.distribution.Distribution):
+        return obj.params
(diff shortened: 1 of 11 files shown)


toslunar:
I checked `examples/gym/train_reinforce_gym.py` runs with and without graphviz.

tests/misc_tests/test_draw_computational_graph.py
```diff
@@ -0,0 +1,103 @@
...
+        vs = chainerrl.misc.collect_variables(v)
+        self.assertEqual(vs, [v])
+
+        vs = chainerrl.misc.collect_variables([v])
+        self.assertEqual(vs, [v])
+
+        vs = chainerrl.misc.collect_variables([[v]])
+        self.assertEqual(vs, [v])
+
+        # ActionValue
```
toslunar: Could you test `QuadraticActionValue`?

chainerrl/misc/draw_computational_graph.py
```diff
@@ -0,0 +1,62 @@
...
+    Args:
+        obj (object): Object to collect Variable objects from.
+    Returns:
+        List of Variable objects.
+    """
+    variables = []
+    if isinstance(obj, chainer.Variable):
+        return [obj]
+    elif isinstance(obj, chainerrl.action_value.ActionValue):
+        return [obj.greedy_actions, obj.max]
```
toslunar: The choice of variables `[obj.greedy_actions, obj.max]` for ActionValue might not be natural to everyone.

muupan: I agree. I chose them because 1) they are required variables of action values and 2) they are sufficient to draw a computational graph. Do you have any suggestions?

chainerrl/misc/is_return_code_zero.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+import subprocess
+
+
+def is_return_code_zero(args):
+    """Return true iff the given command's return code is zero.
+
+    All the messages to stdout or stderr are suppressed.
+    """
+    FNULL = open(os.devnull, 'w')
```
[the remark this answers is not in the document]
muupan: On which environment did you see that warning?

chainerrl/misc/draw_computational_graph.py
toslunar: I would like to add `params` property to `ActionValue` and to postpone drawing `ActionValue` until then.

For example, `greedy_actions` and `max` of `QuadraticActionValue` on an unbounded space are independent of the `mat` argument.

I run
```
qav = chainerrl.action_value.QuadraticActionValue(
    chainer.Variable(np.zeros((5, 5), dtype=np.float32)),
    chainer.Variable(np.ones((5, 5, 5), dtype=np.float32)),
    chainer.Variable(np.zeros((5, 1), dtype=np.float32)),
)

var = chainerrl.misc.collect_variables(qav)
chainerrl.misc.draw_computational_graph(var, 'tmpgraph')
```
- and got
  tmpgraph
- But it should have (5, 5, 5) shaped variable like
  tmpgraph

muupan: Fair enough. I'll implement ActionValue.params.

muupan:
I added ActionValue.params and used them instead of `.max` and `.greedy_actions.`

toslunar:
LGTM

Issue #164, opened 2017-11-07
On gym.undo_logger_setup()


muupan:
Importing `gym` modifies the root logger, which is sometimes undesirable.

ChainerRL depends on `gym` to use `gym.spaces`. Currently, `gym.undo_logger_setup()` is called in `chainerrl/spaces.py` to avoid modyfing the root logger just by using ChainerRL. However, this results in unexpected behavior for users who have been using gym and start trying ChainerRL.

Since gym recently introduced the environment variable `GYM_NO_LOGGER_SETUP` to disable the logger setup globally, now I feel it can be better not to undo the logger setup, but am not fully sure.

In any case, the behavior should be consistent when users import chainerrl. ChainerRL should always import `gym` if it allows `gym` to modify the root logger.

muupan:
If ChainerRL always import gym, it should provide a function like `chainerrl.gym_undo_logger_setup()`.

muupan:
Introducing the `chainerrl.gym` namespace that is not imported from `chainerrl/__init__.py` and moving everything that depends on `gym` under the namespace can be a better option.

By doing so, since users would have to explicitly write `import chainerrl.gym`, dependency of `gym` would become clear.

muupan:
Great news!
https://github.com/openai/gym/commit/4c460ba6c8959dd8e0a03b13a1ca817da6d4074f
> gym.undo_logger_setup is deprecated. gym no longer modifies the global logging configuration

muupan:
Now gym installed via pip no longer modifies the global logging configuration.

PR #167, opened 2017-11-10
[WIP] import gym

toslunar:
#164


toslunar:
This trick doesn't seem to work with Python 2.

PR #171, opened 2017-11-14
Improve MLPBN

muupan:
- Add `nonlinearity` and `last_wscale` argument to MLPBN
- Add tests for MLPBN

The tests for MLPBN are actually not checking each configuration is applied or not, but it should be better than having no tests.

### DIFF
--- a/chainerrl/links/mlp_bn.py
+++ b/chainerrl/links/mlp_bn.py
@@ -29,7 +29,19 @@ class LinearBN(chainer.Chain):
 
 
 class MLPBN(chainer.Chain):
-    """Multi-Layer Perceptron with BatchNormalization."""
+    """Multi-Layer Perceptron with Batch Normalization.
+
+    Args:
+        in_size (int): Input size.
+        out_size (int): Output size.
+        hidden_sizes (list of ints): Sizes of hidden channels.
+        normalize_input (bool): If set to True, Batch Normalization is applied
+            to inputs.
+        normalize_output (bool): If set to True, Batch Normalization is applied
+            to outputs.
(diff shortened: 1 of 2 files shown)


chainerrl/links/mlp_bn.py
```diff
@@ -27,15 +29,28 @@ def __call__(self, x):
...
+
+    Args:
+        in_size (int): Input size.
+        out_size (int): Output size.
+        hidden_sizes (list of ints): Sizes of hidden channels.
+        normalize_input (bool): If set to True, Batch Normalization is applied
+            to inputs.
+        normalize_output (bool): If set to True, Batch Normalization is applied
+            to outputs.
+        nonlinearity (callable): Nonlinearity between layers.
```
toslunar: In most cases, `nonlinearity` is a `chainer.Function`, and it cannot be a `chainer.Link`.  Could you make the followings clear?
- `nonlinearity` has no parameters;
- `nonlinearity`, called with one argument of `chainer.Variable`, returns `chainer.Variable` of the same shape as the input.

tests/links_tests/test_mlp_bn.py
```diff
@@ -0,0 +1,60 @@
...
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from builtins import *  # NOQA
+from future import standard_library
+standard_library.install_aliases()
+
+import unittest
+
+import numpy as np
```
toslunar: In other files, `numpy` and `chainer` are in the same group of the import order.

toslunar:
LGTM

PR #172, opened 2017-11-15
Improve StateActionQFunctions

muupan:
Please merge this after merging #171 

- Add `nonlinearity` and `last_wscale` aruguments to StateActionQFunctions
- Improve docstrings
- Add tests for StateActionQFunctions. They are actually not checking if each configuration is applied or not, but it should be better than having no test.

### DIFF
--- a/chainerrl/q_functions/state_action_q_functions.py
+++ b/chainerrl/q_functions/state_action_q_functions.py
@@ -42,7 +42,10 @@ class FCSAQFunction(MLP, StateActionQFunction):
         n_dim_action (int): Number of dimensions of action space.
         n_hidden_channels (int): Number of hidden channels.
         n_hidden_layers (int): Number of hidden layers.
-        nonlinearity (callable): Nonlinearity between layers.
+        nonlinearity (callable): Nonlinearity between layers. It must accept a
+            Variable as an argument and return a Variable with the same shape.
+            Nonlinearities with learnable parameters such as PReLU are not
+            supported.
         last_wscale (float): Scale of weight initialization of the last layer.
     """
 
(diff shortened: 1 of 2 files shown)


chainerrl/q_functions/state_action_q_functions.py
```diff
@@ -130,25 +148,34 @@ def __call__(self, state, action):
...
+                 n_hidden_layers, normalize_input=True,
+                 nonlinearity=F.relu, last_wscale=1.):
         self.n_input_channels = n_dim_obs + n_dim_action
         self.n_hidden_layers = n_hidden_layers
         self.n_hidden_channels = n_hidden_channels
         self.normalize_input = normalize_input
+        self.nonlinearity = nonlinearity
 
         super().__init__()
         with self.init_scope():
```
toslunar: Could you add a comment that `nonlinearity` does not need to be passed to `MLPBN` because `hidden_sizes` is empty?

tests/q_functions_tests/test_state_action_q_function.py
```diff
@@ -0,0 +1,201 @@
...
+    @attr.gpu
+    def test_call_gpu(self):
+        self._test_call(gpu=0)
+
+
+@testing.parameterize(
+    *testing.product({
+        'n_dim_obs': [1, 5],
+        'n_dim_action': [1, 3],
+        'n_hidden_layers': [0, 1, 2],
```
toslunar: `'n_hidden_layers': [1, 2],`

muupan: n_hidden_layers=0 works except for LateAction architectures. I'll remove 0 for them.

muupan:
Thank you for your review! I fixed the points you mentioned.

PR #182, opened 2017-11-22
Patches to envs.ALE

toslunar:
- Change grayscale formula to `0.299 R' + 0.587 G' + 0.114 B'`, where the values are gamma compressed.
- Fix `frame_skip` argument.
- Add and fix assertions.

### DIFF
--- a/chainerrl/envs/ale.py
+++ b/chainerrl/envs/ale.py
@@ -116,10 +116,9 @@ class ALE(env.Env):
         rgb_img = np.maximum(self.ale.getScreenRGB(), self.last_raw_screen)
         # Make sure the last raw screen is used only once
         self.last_raw_screen = None
-        assert rgb_img.shape == (210, 160, 3)
+        assert rgb_img.shape[2:] == (3,)
         # RGB -> Luminance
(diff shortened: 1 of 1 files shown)


chainerrl/envs/ale.py
```diff
@@ -116,10 +118,9 @@ def current_screen(self):
         rgb_img = np.maximum(self.ale.getScreenRGB(), self.last_raw_screen)
         # Make sure the last raw screen is used only once
         self.last_raw_screen = None
-        assert rgb_img.shape == (210, 160, 3)
```
muupan: Screen sizes can be different from (210, 160) when you use wrong rom files, so it's better to check them.

muupan:
Good catch for the grayscale formula. It seems like deepmind and openai use your formula, and so should we.

chainerrl/envs/ale.py
toslunar: This change removes a double-check.  This line checks rgb-format and L125--127 checks screen sizes.

muupan: I see.

PR #186, opened 2017-11-26
[WIP] Add Quantile Regression DQN

toslunar:
http://arxiv.org/abs/1710.10044


HassamSheikh:
Is it complete?

toslunar:
Thanks for your having an interest in this PR. The current status is
- one can run the agent `QRDQN`;
- I'm sorry if there's a bug since I've not checked its performance;
- `_ActionQuantile` may be available later in `chainerrl.action_value`.

PR #189, opened 2017-11-30
Added agent test for act without train (e.g. batch normalization)

rkawajiri:
below agents seem to be fixed

- [x] a3c
- [x] acer
- [x] reinforce


muupan:
Since the current CI outputs for this PR contain exceptions raised in a3c tests  (#193), please don't merge this even if CI looks ok. The exceptions should be resolved before merging this.

rkawajiri:
In discussion with owners, we realized that a3c or other algorithm cannot be essentially used with batch normalization. So this PR is unnecessary and I'm closing.

PR #190, opened 2017-11-30
Fix QuadraticActionValue and add tests

toslunar:
`QuadraticActionValue` called with `min_action=None` or `max_action=None` seems to have a bug.

Note:
```
>>> import numpy as np
>>> np.asarray(None, dtype=np.float32)
array(nan, dtype=float32)
```

### DIFF
--- a/tests/test_action_value.py
+++ b/tests/test_action_value.py
@@ -59,3 +59,40 @@ class TestDiscreteActionValue(unittest.TestCase):
                 v = self.q_values[b, greedy_actions[b]]
                 adv = q - v
                 self.assertAlmostEqual(ret.data[b], adv)
+
+
+class TestQuadraticActionValue(unittest.TestCase):
+    def test_unbounded(self):
+        n_batch = 7
+        ndim_action = 3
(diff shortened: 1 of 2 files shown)


tests/test_action_value.py
```diff
@@ -59,3 +59,40 @@ def test_compute_advantage(self):
...
+        mu_is_allowed = (min_action < mu) * (mu < max_action)
+        mu_is_allowed = np.all(mu_is_allowed, axis=1)
+
+        # If mu[i] is an valid action, v_out[i] should be v[i]
+        np.testing.assert_almost_equal(v_out[mu_is_allowed], v[mu_is_allowed])
+
+        # Otherwise, v_out[i] should be less than v[i]
+        np.testing.assert_array_less(
+            v_out[~mu_is_allowed],
+            v[~mu_is_allowed] + 1e-4)
```
muupan: How about asserting `v_out[~mu_is_allowed]` is strictly smaller than `v[~mu_is_allowed]`, by removing `+ 1e-4`? The current assertion doesn't fail if they have the same values i.e. min_action and max_action are not used at all.

muupan:
LGTM

PR #194, opened 2017-12-01
Warn about negative exit code of child processes

muupan:


chainerrl/misc/async.py
```diff
@@ -132,6 +133,10 @@ def set_seed_and_run(process_idx, run_func):
 
     for p in processes:
         p.join()
+        if p.exitcode < 0:
+            warnings.warn(
+                "Process {}'s exit code was not zero but {}".format(
```
toslunar: I'd like to know the process_idx.
How about `"Process #{} (pid={}) exited with nonzero status {}".format(i, p.pid, p.exitcode)`?

muupan:
I added a test to check if warnings are raised or not.

PR #201, opened 2017-12-12
Raise appropriate errors if protocols of PriorityReplayBuffer are not followed

toslunar:
I suppose `assert` is not user-friendly enough in these cases.
Besides the PR adds tests.


toslunar:
Just refactor the test at this moment. (#218)
`RuntimeError` seems no better than `AssertionError`.

PR #204, opened 2017-12-16
TRPO agent

### LINKED ISSUE #202: Trust Region Policy Optimization (TRPO)
- Newer versions of Chainer support double backprop
- TRPO is still one of the state-of-the-art algorithms for continuous-action domains

muupan:
This PR adds `chainerrl.agents.TRPO`, which implements the TRPO-GAE algorithm.

Resolves #202 

~~Strangely, chainer==3.0.0 and chainer==3.1.0 have different results. With chainer==3.0.0 line search always backtracks several times, while with chainer==3.1.0 it rarely backtracks. Values of expected improve and KL div are different.~~ This problem is solved. Current TRPO only works with 3.1.0 or later.

`python examples/gym/train_trpo_gym.py --env Hopper-v1 --gpu -1`

chainer==3.0.0
```
INFO:chainerrl.experiments.train_agent:outdir:results/20171217T050456.738713 step:1022 episode:48 R:9.23186813481
INFO:chainerrl.experiments.train_agent:statistics:[('average_value', nan), ('average_entropy', 3.7070460319519043), ('average_kl', nan), ('average_policy_step_size', nan)]
INFO:chainerrl.agents.trpo:Line search iteration: 0 step size: 1.0
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.13816076517105103
INFO:chainerrl.agents.trpo:KL divergence: 0.8066530227661133
INFO:chainerrl.agents.trpo:KL divergence exceeds max_kl. Bakctracking...
INFO:chainerrl.agents.trpo:Line search iteration: 1 step size: 0.5
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.08148320019245148
INFO:chainerrl.agents.trpo:KL divergence: 0.20232915878295898
INFO:chainerrl.agents.trpo:KL divergence exceeds max_kl. Bakctracking...
INFO:chainerrl.agents.trpo:Line search iteration: 2 step size: 0.25
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.04137067496776581
INFO:chainerrl.agents.trpo:KL divergence: 0.05066625028848648
INFO:chainerrl.agents.trpo:KL divergence exceeds max_kl. Bakctracking...
INFO:chainerrl.agents.trpo:Line search iteration: 3 step size: 0.125
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.020503554493188858
INFO:chainerrl.agents.trpo:KL divergence: 0.012677105143666267
INFO:chainerrl.agents.trpo:KL divergence exceeds max_kl. Bakctracking...
INFO:chainerrl.agents.trpo:Line search iteration: 4 step size: 0.0625
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.010174408555030823
INFO:chainerrl.agents.trpo:KL divergence: 0.003170591313391924
```

chainer==3.1.0
```
INFO:chainerrl.experiments.train_agent:outdir:results/20171217T050602.057921 step:1022 episode:48 R:9.23186813481
INFO:chainerrl.experiments.train_agent:statistics:[('average_value', nan), ('average_entropy', 3.7070460319519043), ('average_kl', nan), ('average_policy_step_size', nan)]
INFO:chainerrl.agents.trpo:Line search iteration: 0 step size: 1.0
INFO:chainerrl.agents.trpo:Surrogate objective improve: 0.03503759950399399
INFO:chainerrl.agents.trpo:KL divergence: 0.005451631732285023
```


muupan:
I noticed 3.1.0 adds double-backprop support for relevant functions such as softplus and log. It seems, unfortunately, that calling `chainer.grad` twice for functions that don't support double-backprop doesn't raise an error.

code
```
import chainer
from chainer import functions as F
import numpy as np
x = chainer.Variable(np.zeros((1, 1), dtype=np.float32))
y = F.softplus(x)
g = chainer.grad([y], [x], enable_double_backprop=True)
gg = chainer.grad([g[0]], [x])
print('g', g)
print('gg', gg)
```

3.0.0
```
g [variable([[ 0.5]])]
gg [None]
```

3.1.0
```
g [variable([[ 0.5]])]
gg [variable([[ 0.25]])]
```

This explains different TRPO results.

muupan:
I think it should raise an error if computation contains functions that don't support double-backprop, but am not sure how to detect them.

`None` returned by `chainer.grad` doesn't always mean existence of such functions, because double-backrop of `y = 2 * x` is `None`, too.

muupan:
I added `_find_old_style_function` to detect old-style functions and check it before double backprop. From Chainer v3.1.0 there's no old-style functions in TRPO's tests and example.

muupan:
TODO:
- normalize observations
- compare performance to openai/baselines

muupan:
python examples/gym/train_trpo_gym.py --env Hopper-v1 --steps 2000000 --eval-interval 100000 --eval-n-runs 100
trpo_hopper

python examples/gym/train_trpo_gym.py --env Walker2d-v1 --steps 2000000 --eval-interval 100000 --eval-n-runs 100
trpo_walker2d

While these are single runs from random seed 0, their performance looks better than those in the PPO paper, and comparable to http://arxiv.org/abs/1709.06560 as well.

tests/agents_tests/test_trpo.py
```diff
@@ -0,0 +1,385 @@
...
+            self.skipTest("""\
+Chainer v{} does not support double backprop of these functions: {}.""".format(
+                chainer.__version__, old_style_funcs))
+
+        def test_hessian_vector_product_zero(vec):
+            hvp = compute_hessian_vector_product(y, params, vec)
+            hessian = compute_hessian(y, params)
+            self.assertEqual(np.count_nonzero(hvp), 0)
+            self.assertEqual(np.count_nonzero(hessian), 0)
+            np.testing.assert_allclose(hvp, hessian.dot(vec), atol=1e-3)
```
toslunar: This line seems equivalent to checking shapes, because 0 × 0 = 0.

chainerrl/agents/trpo.py
```diff
@@ -0,0 +1,572 @@
...
+        self.act_deterministically = act_deterministically
+        self.logger = logger
+
+        self.value_record = collections.deque(maxlen=value_stats_window)
+        self.entropy_record = collections.deque(maxlen=entropy_stats_window)
+        self.kl_record = collections.deque(maxlen=kl_stats_window)
+        self.policy_step_size_record = collections.deque(
+            maxlen=policy_step_size_stats_window)
+
+        self.xp = self.policy.xp
```
toslunar: If it's allowed to put `policy` and `vf` on different devices (cpu / gpu), `self.xp` might be confusing.

chainerrl/policies/gaussian_policy.py
```diff
@@ -145,13 +145,20 @@ class FCGaussianPolicyWithStateIndependentCovariance(
...
 
     def __init__(self, n_input_channels, action_size,
                  n_hidden_layers=0, n_hidden_channels=None,
                  min_action=None, max_action=None, bound_mean=False,
                  var_type='spherical',
-                 nonlinearity=F.relu, mean_wscale=1):
+                 nonlinearity=F.relu,
+                 mean_wscale=1,
+                 var_func=F.softplus,
+                 var_param_init=0,
```
toslunar: Do you think the name `var_param_init` suggests the correspondence to the names `var_wscale` and `var_bias` of `FCGaussianPolicy`?

muupan: Maybe `var` is more consistent with `var_bias`, while less informative. Do you think `var` is better?

chainerrl/agents/trpo.py
```diff
@@ -0,0 +1,572 @@
...
+def _replace_params_data(params, new_params_data):
+    """Replace data of params with new data."""
+    for param, new_param_data in zip(params, new_params_data):
+        assert param.shape == new_param_data.shape
+        param.data[:] = new_param_data
+
+
+def _hessian_vector_product(flat_grads, params, vec):
+    """Compute hessian vector product efficiently by backprop."""
+    grads = _chainer_grad_with_zero([F.sum(flat_grads * vec)], params)
```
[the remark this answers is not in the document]
muupan: Right. We cannot assume that for general hessian vector product, but it's true TRPO doesn't work with unused parameters because of CG. I think `chainer.grad` and raising an informative error when there's None would be better. I'll fix it.

chainerrl/policies/gaussian_policy.py
muupan: Actually, this param does not represent variance, but it represents values that are converted to variance via `var_func`, so I added `_param`. I admit it is still confusing, but I didn't come up with a better name. Any suggestion?

muupan:
I fixed the points you mentioned, except the name of `var_param_init`, which I don't come up a better name.

muupan:
The coverage with Chainer v2 decreased while that of v3 increased. I think it is because the tests of TRPO are active only for v3.

PR #210, opened 2017-12-28
show error message

corochann:
When we construct `QFunction`, and model parameter is not initialized (it happens when some Link is instantiated with `in_size=None`), `copy_param` will fail because the `target_link`'s `param` is `None`.

This PR is to show user-friendly error message to let the user know what is the cause of this error.

I'd like to get comment that what kind of message is better.

I'm not sure the performance degrade by checking this type error.

### DIFF
--- a/chainerrl/misc/copy_param.py
+++ b/chainerrl/misc/copy_param.py
@@ -16,7 +16,9 @@ def copy_param(target_link, source_link):
         if target_params[param_name].data is None:
             raise TypeError(
                 'target_link parameter {} is None. Maybe the model params are '
-                'not initialized.'.format(param_name))
+                'not initialized.\nPlease try to forward dummy input beforehand'
+                ' to determine parameter shape of the model.'.format(
+                    param_name))
         target_params[param_name].data[:] = param.data
 
     # Copy Batch Normalization's statistics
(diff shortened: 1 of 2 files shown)


toslunar:
Thanks. I agree with you about showing error message.

I'd like to see message that starts with `"target_link has parameter {}"`, since `target_params` is a local name.

corochann:
And should we add some sentence to how to fix this error like:

'target_link parameter {} is None. Maybe the model params are not initialized.\n Please forward dummy input before hand to determine parameter shape.'.format(param_name)

or something?

toslunar:
It's a possible fix to forward dummy input.  However, I'd rather not recommend it.
Instead default agents in ChainerRL compute the first action before doing the first training (in `act_and_train`).

corochann:
Updated text.

> default agents in ChainerRL compute the first action before doing the first training (in act_and_train).

Yeah, if `DQN` agent uses parameter-initialized `target_link` it would be nice.

toslunar:
Sorry.  You are right.  `DQN` agent expects parameters of `q_function` to be initialized.  Let's recommend forwarding dummy input on such uninitialized-parameter errors.

corochann:
So you prefer below comment?

`'target_link parameter {} is None. Maybe the model params are not initialized.\n Please forward dummy input before hand to determine parameter shape of the model.'.format(param_name)`

toslunar:
Yes, I prefer the longer comment.

Some minor fixes:
- `\n Please` -> `\nPlease`
- `Please` -> `Please try to` (?)
- `before hand` -> `beforehand`

toslunar:
Could you add testcases (in `tests/misc_tests/test_copy_param.py`) if you could spare the time?

PR #214, opened 2018-01-09
[WIP] Fix on flake8 3.5.0

toslunar:
The current check misses some new rules.
> "flake8 with hacking forces flake8 to be old (2.5.5) that uses pep8" https://github.com/chainer/chainer/pull/4097#issuecomment-351624573


toslunar:
Fixed the current `master` except that I left many lines with the following lints: `E741 ambiguous variable name 'l'`.

PR #215, opened 2018-01-10
Make tests check exit status of subprocesses

toslunar:
- Warn nonzero exit status. (see also #194)
- Check warnings in tests using `train_agent_async`.


muupan:
LGTM

PR #216, opened 2018-01-10
make ReplayBuffer.load() compatible with v0.2.0.

mr4msm:
If we use ReplayBuffer that loaded a buffer saved with v0.2.0, an error will occur because deque doesn't have sample() method.
This PR makes load() methods of ReplayBuffer and EpisodicReplayBuffer compatible with buffer saved with v0.2.0.

### DIFF
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -10,6 +10,7 @@ standard_library.install_aliases()
 from abc import ABCMeta
 from abc import abstractmethod
 from abc import abstractproperty
+import collections
 
 import numpy as np
 import six.moves.cPickle as pickle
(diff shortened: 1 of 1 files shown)


chainerrl/replay_buffer.py
```diff
@@ -153,6 +153,9 @@ def save(self, filename):
     def load(self, filename):
         with open(filename, 'rb') as f:
             self.memory = pickle.load(f)
+        if not isinstance(self.memory, RandomAccessQueue):
```
toslunar: `PrioritizedReplayBuffer.load` doesn't work well, because the class extends `ReplayBuffer` and its `memory` is not `RandomAccessQueue`.

toslunar: `PrioritizedReplayBuffer.load` doesn't work well, because the class extends `ReplayBuffer` and its `memory` is not `RandomAccessQueue`.

mr4msm: Thanks for your comment.
Sorry, I completely forgot to consider prioritized buffers.

mr4msm:
I appreciate your comment, additional commit and PR #217.

PR #222, opened 2018-01-29
Add requirements-dev.txt

muupan:
to make easy to set up a development environment.


toslunar:
The current test requires `atari-py` and `mock`.

muupan:
Thank you. I added them.

PR #226, opened 2018-02-05
gym 0.9.6 is not working with python2

toslunar:


muupan:
Can you modify https://github.com/chainer/chainerrl/blob/master/setup.py accordingly?

toslunar:
Done

muupan:
LGTM

toslunar:
FYI, gym 0.9.7 works with Python 2.

PR #234, opened 2018-02-13
Add docstring to ALE and fix seed range

muupan:


muupan:
Thank you, I fixed it.

PR #239, opened 2018-02-14
Improve random seed setting in ALE examples

muupan:
Merge #234 first.

This PR improves default random seed settings in `examples/ale/train_dqn_ale.py` and `examples/ale/train_a3c_ale.py` so that they are deterministic by default.

If this is merged, then I'll update other examples as well in another PR.

### DIFF
--- a/examples/ale/train_a3c_ale.py
+++ b/examples/ale/train_a3c_ale.py
@@ -97,8 +97,8 @@ def main():
     # Set different random seeds for different subprocesses.
     # If seed=0 and processes=4, subprocess seeds are [0, 1, 2, 3].
     # If seed=1 and processes=4, subprocess seeds are [4, 5, 6, 7].
-    assert (args.seed + 1) * args.processes < 2 ** 31
     process_seeds = np.arange(args.processes) + args.seed * args.processes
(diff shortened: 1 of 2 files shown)


examples/ale/train_a3c_ale.py
```diff
@@ -88,11 +89,18 @@ def main():
...
+    # Set a random seed used in ChainerRL.
+    # If you use more than one processes, the results will be no longer
+    # deterministic even with the same random seed.
+    misc.set_random_seed(args.seed)
 
-    args.outdir = experiments.prepare_output_dir(args, args.outdir)
+    # Set different random seeds for different subprocesses.
+    # If seed=0 and processes=4, subprocess seeds are [0, 1, 2, 3].
+    # If seed=1 and processes=4, subprocess seeds are [4, 5, 6, 7].
+    assert (args.seed + 1) * args.processes < 2 ** 31
```
toslunar: The check would be `assert (args.seed + 1) * args.processes - 1 < 2 ** 31`.  It might be clearer to check it by `process_seeds.max() < 2 ** 31`.

muupan:
Thank you for the review. I fixed the boundary check.

Issue #240, opened 2018-02-14
Use backward immediately after loss computation to increase memory efficiency


lyx-x:
This is a general remark after I got memory issue with PCL on Atari games. I noticed that losses are sometimes appended into a list (as shown below) and a weighted loss is computed (much) later. This is in theory correct, but memory inefficient as Chainer retrains all intermediate computation result in memory, e.g. when we apply a network on 2 inputs, there will be 2 copies of intermediate results and 2 losses.

https://github.com/chainer/chainerrl/blob/4b051a4cc0e13634b9b0f521c0ad18db91019d40/chainerrl/agents/pcl.py#L221

Since Functions and Links accumulates gradients and gradients are linearly additives, I would suggest we call backward immediately after getting a loss instead of saving them to a list. The cleargrads can be called in some high level function such as act_and_train.

Issue #238, opened 2018-02-14
PCL's docstring is totally wrong


muupan:
For example, It uses A3CModel, not ACERModel.

muupan:
Seems already solved, maybe my mistake.

Issue #236, opened 2018-02-14
[Bug] Argument t_max in PCL misused


lyx-x:
According to the docstring below, `t_max` in PCL is equivalent to `update_interval` in DQN.

https://github.com/chainer/chainerrl/blob/4ca100cd8f0dab5e5e333cf3bdc65fb6e6e30651/chainerrl/agents/pcl.py#L49

However, if `t_max == None` by default (also in the example script), line 388 (below), the update, will never be executed. 

https://github.com/chainer/chainerrl/blob/4b051a4cc0e13634b9b0f521c0ad18db91019d40/chainerrl/agents/pcl.py#L388

In addition, when sampling from the replay buffer, I think that `max_len` should be set with another parameter, at least not with `t_max` unless we change the docstring.  (I also think there is no such limit on the length of episode in the original paper)

https://github.com/chainer/chainerrl/blob/4b051a4cc0e13634b9b0f521c0ad18db91019d40/chainerrl/agents/pcl.py#L284

muupan:
When `t_max` is set to None, PCL is expected to work as the original paper does, i.e. sampling a full episode.
When `t_max` is set to non-None value, it is expected to work like ACER with PCL's loss function.

It seems the docstring of PCL is wrong, not only in `t_max` but also in other points. We should fix this. Thank you for pointing it out.

Issue #109, opened 2017-06-12
Unified PCL


ethancaballero:
See section 5.1 for new more performant update to PCL: https://arxiv.org/pdf/1702.08892.pdf

lyx-x:
The unified version is already supported by the current implementation (at least in theory). The idea is to use a SharedModel for Q value and 2 heads (without trainable parameters) for pi and V using the formula in the paper.

I believe we can close this issue.

PR #245, opened 2018-02-25
Rewrite PCL agent

lyx-x:
I rewrote the PCL agent to avoid memory issues when saving Variables inside list / replay buffer. I didn't compare the training curve with the old one, but it seems to learn (the average_value increases and R gets bigger) on Catpole under the new parameters and there is no memory issue when run with large network / reasonably long trajectories. 

Main methods are the following:

`update`: take a loss (as an array), log the result as usual and call optimizer (the backprop is done before this function is called)
`update_on_policy` and `update_from_replay`: sample a list of trajectories (from replay or the current one), clear grads and compute loss
`compute_loss`: take a list of trajectories, perform batch computation (batch size is the number of episodes, which may not be efficient when there is one single episode for on-policy update). This function will call backward immediately and only return an array for logging
`_compute_path_consistency`: compute path consistency, this part of code is almost unchanged

The new underlying data structure is a list of dict to store the current episode, then a replay buffer that only stores (s,a,r) pairs. The old mu (action_distrib) is removed since it can be recomputed again from other items.

I also added a unified model in the example script and changed a couple of parameters.

Issues addressed: #109 #236 #240 

I am not sure if the parameters are used correctly, but if they are correct, this PR also addresses #238


muupan:
Thank you for the improvements on PCL. I haven't checked the implementation details yet, but I think solving the memory issue is great as long as it won't make training slow.

Can you show the training curves and computation speeds before and after this PR?

PR #247, opened 2018-03-15
Run autopep8 in travis ci

muupan:


muupan:
Added autopep8 to requirements-dev.txt.

PR #249, opened 2018-03-16
Categorical DQN (C51)

muupan:
- [x] Check performance on Atari
- [x] Add tests
  - [x] tests of C51
  - [x] tests of DistributionalDiscreteActionValue
- [x] Clean code
  - [x] ~Implement its own `__init__`~ Clarify differences from DQN in the docstring

~Merge #248 first.~


muupan:
Below are the results of `examples/ale/train_c51_ale.py {rom}` and `examples/ale/train_dqn_ale {rom} --agent DQN/DoubleDQN/PAL`. C51 achieves better scores across games than other DQN variants.

asterix
beam_rider
breakout
pong
qbert
seaquest
space_invaders

muupan:
It is unclear to me that categorical projection should be implemented as Algorithm 1 or as (7) in the paper. Algorithm 1 seems wrong to me when b_j is an integer.

chainerrl/agents/categorical_dqn.py
```diff
@@ -0,0 +1,138 @@
...
+            batch_q_target = self._compute_target_values(exp_batch, gamma)
+            assert batch_q_target.shape == (batch_size, n_atoms)
+
+        return batch_q, batch_q_target
+
+    def _compute_loss(self, exp_batch, gamma, errors_out=None):
+        """Compute a loss of categorical DQN."""
+        y, t = self._compute_y_and_t(exp_batch, gamma)
+        # minimize the cross entropy
+        eltwise_loss = -t * F.log(F.clip(y, 1e-10, 1.))
```
toslunar: Could you explain why `F.clip` is here?

chainerrl/agents/categorical_dqn.py
```diff
@@ -0,0 +1,138 @@
...
+        scatter_add = np.add.at
+
+    z_probs = xp.zeros((batch_size, n_atoms), dtype=xp.float32)
+    offset = xp.arange(
+        0, batch_size * n_atoms, n_atoms, dtype=xp.int32)[..., None]
+    # Accumulate m_l
+    scatter_add(
+        z_probs.ravel(),
+        (l.astype(xp.int32) + offset).ravel(),
+        (y_probs * (u - bj)).ravel())
```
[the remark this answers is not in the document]
muupan: Wow, your solution is definitely better than mine!

chainerrl/agents/categorical_dqn.py
muupan: I found clipping was necessary for training CategoricalDQN from earlier experiments. Without clipping, some probability values converges to 0, resulting in log(0) -> NaN.

Other unofficial implementations also apply clipping.
https://github.com/Kaixhin/Rainbow/blob/master/agent.py#L85
https://github.com/floringogianu/categorical-dqn/blob/master/policy_improvement/categorical_update.py#L53

Since clipping by 1e-10 worked, I didn't tune 1e-10 further. It is possible larger values may result in better performance.

muupan: I added a comment to explain why.

muupan:
Thanks for the review. I fixed them.

Issue #268, opened 2018-05-13
How do we get the rom for ale examples?


arisliang:
Do we need to download the rom and specify it ourselves? Do we have any examples?

toslunar:
No, we don't have to.  `atari_py` contains rom data.  Could you try the following?
```
>>> import atari_py  # 0.1.1
>>> ale = atari_py.ALEInterface()
>>> ale.loadROM(atari_py.get_game_path('Adventure'))
>>> import matplotlib.pyplot as plt
>>> plt.imshow(ale.getScreenRGB())
<matplotlib.image.AxesImage object at 0x110f2dc88>
>>> plt.show()
```

arisliang:
error at 3rd line `ale.loadROM(atari_py.get_game_path('Adventure'))`
> No ROM File specified or the ROM file was not found.

But I tried other game such as `python train_dqn_ale.py breakout` worked.

arisliang:
The above lines works for lower case `adventure`, however training script `python train_dqn_ale.py adventure` doesn't work. It gives this error:
> This ROM is for PAL. Please use ROMs for NTSC

toslunar:
> RuntimeError: This ROM is for PAL. Please use ROMs for NTSC

This error is on the screen size.  The default preprocessing supports only images with a certain size.

muupan:
> RuntimeError: This ROM is for PAL. Please use ROMs for NTSC

Sorry, this is kind of a bug of the current `chainerrl.envs.ale.ALE`. Some of the ROMs in atari_py do not work because of this. To alleviate this problem, you can use your own NTSC ROM or use https://github.com/chainer/chainerrl/pull/253.

muupan:
BTW it seems `gym` does not support Adventure while atari_py supports it.

PR #253, opened 2018-04-06
[WIP] Use gym and atari wrappers instead of chainerrl.envs.ale

muupan:
- [ ] Compare performance of agents
- [ ] Compare speed

PR #269, opened 2018-05-14
Set ROM name to lowercase as expected in atari_py

arisliang:
Issue mentioned in #268


muupan:
I think this PR is good, but since we are planning to switch to using Atari env implementations in `gym` #253 and stop directly specifying rom names, I keep this PR unmerged for now. Thanks anyway for your contribution.

PR #271, opened 2018-05-16
Remove unnecessary declaration and substitution of 'done' in the train_agent function 

uidilr:
The declaration and substitution of 'done' in the train_agent function is unnecessary because 'done' is created and substituted by env.step and it has proper value for training.

The results of pytest with gpu is:
```
 6124 passed, 147 warnings in 1592.54 seconds 
```

Update of .gitignore is included in this PR


muupan:
Good point. Thanks for your contribution.

PR #277, opened 2018-06-04
FIFO prioritized replay buffer

toslunar:
- pop the earliest experience when it's full
- return true probability (except for the effect of non-repetitive sampling) when the sampling is mixed with uniform sampling


muupan:
LGTM

PR #280, opened 2018-06-19
Update Policy class to inherit ABCMeta 

uidilr:
Updated Policy class to inherit ABCMeta. 
It is because Policy class had abstract method, but did not inherit ABCMeta class.
[Python docs](https://docs.python.org/3/library/abc.html) says that decorator `@abstractmethod` requires that the class’s metaclass is ABCMeta or is derived from it.


muupan:
Thank you for the PR. I think it should be merged once https://github.com/chainer/chainerrl/issues/281 is solved.

muupan:
👍

PR #286, opened 2018-07-08
Modify `async` to `async_` to support Python 3.7

mmilk1231:
`async` becomes a keyword in Python 3.7. Thus when I import ChainerRL with Python 3.7, the following SyntaxError happened:

```
  File "/usr/local/lib/python3.7/site-packages/chainerrl-0.3.0-py3.7.egg/chainerrl/agents/a3c.py", line 122
    async.assert_params_not_shared(self.shared_model, self.model)
         ^
SyntaxError: invalid syntax
...
```

This PR avoid the error by modifying `async` to `async_`.

References about Python 3.7:
https://www.python.org/dev/peps/pep-0492/
https://docs.python.org/3/whatsnew/3.7.html#changes-in-python-behavior


chainerrl/agents/nsq.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, q_function, optimizer,
         self.target_q_function = copy.deepcopy(q_function)
         self.q_function = copy.deepcopy(self.shared_q_function)
 
-        async.assert_params_not_shared(self.shared_q_function, self.q_function)
+        async_.assert_params_not_shared(self.shared_q_function, self.q_function)
```
toslunar: Could you break this line (>79 chars)?

mmilk1231: I'm sorry for my mistakes. I break this line in the new commit.

mmilk1231:
I fixed `chainerrl/tests/misc_tests/test_async.py` by rewriting `async` to `async_`.
Thank you for pointing it out.

PR #287, opened 2018-07-18
Noisy network fixes

seann999:


chainerrl/links/noisy_linear.py
```diff
@@ -26,10 +27,16 @@ def __init__(self, mu_link, sigma_scale=0.4):
 
         with self.init_scope():
             self.mu = mu_link
+            self.mu.W.initializer = Uniform(1 / numpy.sqrt(in_size))
+            if not self.nobias:
+                self.mu.b.initializer = Uniform(1 / numpy.sqrt(in_size))
+
+            self.mu.W.initialize((self.out_size, in_size))
+            self.mu.b.initialize((self.out_size))
```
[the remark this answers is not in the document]
seann999: The original paper stated that mu was initialized with U[-1/sqrt(fan_in), 1/sqrt(fan_in)]. Should I instead add initialization options to links.dqn_head as I think you suggested?

chainerrl/links/noisy_linear.py
```diff
@@ -26,10 +27,16 @@ def __init__(self, mu_link, sigma_scale=0.4):
...
             self.sigma = L.Linear(
                 in_size=in_size, out_size=self.out_size, nobias=self.nobias,
                 initialW=VarianceScalingConstant(sigma_scale),
-                initial_bias=Constant(sigma_scale))
+                initial_bias=Constant(sigma_scale / numpy.sqrt(self.out_size)))
```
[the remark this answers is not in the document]
seann999: I see. Empirically, there also wasn't a big performance difference so I'll change it to `sqrt(in_size)`.

chainerrl/links/noisy_linear.py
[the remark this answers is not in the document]
toslunar: I didn't think it matters much whether we use normal or uniform, but I've just noticed U[-1/sqrt(fan_in), 1/sqrt(fan_in)] is \sqrt{3} or \sqrt{6} times small than usual!  Therefore modifying `mu_link` here makes sense.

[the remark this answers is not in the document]
toslunar: Why don't you fix `__call__` to support lazy initialization of the parameter `W`?

chainerrl/links/noisy_linear.py
```diff
@@ -26,10 +28,22 @@ def __init__(self, mu_link, sigma_scale=0.4):
 
         with self.init_scope():
             self.mu = mu_link
+            self.mu.W.initializer = VarianceScalingUniform(1)
+            if not self.nobias:
+                self.mu.b.initializer = VarianceScalingUniform(1)
+
+            # assure weight reinitialization
```
toslunar: I feel reinitialization is tricky.  How about creating a new `L.Linear` for `self.mu` (like the interface before the commit https://github.com/chainer/chainerrl/pull/169/commits/6af88d63c97b0b6e36eedc5c7e03190ec314cb52#diff-3d0a1daaad7846c7235e591fbf316217)?

chainerrl/links/noisy_linear.py
```diff
@@ -26,10 +28,22 @@ def __init__(self, mu_link, sigma_scale=0.4):
 
         with self.init_scope():
             self.mu = mu_link
+            self.mu.W.initializer = VarianceScalingUniform(1)
+            if not self.nobias:
+                self.mu.b.initializer = VarianceScalingUniform(1)
```
toslunar: The bias parameter `b` is usually initialized as `0`.  On `mu.b`, don't you think it's also reasonable?

chainerrl/links/noisy_linear.py
```diff
@@ -26,10 +28,22 @@ def __init__(self, mu_link, sigma_scale=0.4):
 
         with self.init_scope():
             self.mu = mu_link
+            self.mu.W.initializer = VarianceScalingUniform(1)
```
toslunar: `chainer.initializers.LeCunUniform(1/np.sqrt(3))` can be used.

seann999: I never thought of that!

chainerrl/links/noisy_linear.py
seann999: That is true.

chainerrl/links/noisy_linear.py
```diff
@@ -25,11 +25,13 @@ def __init__(self, mu_link, sigma_scale=0.4):
...
-                in_size=in_size, out_size=self.out_size, nobias=self.nobias,
-                initialW=VarianceScalingConstant(sigma_scale),
-                initial_bias=Constant(sigma_scale))
+            self.mu = L.Linear(in_size, self.out_size, self.nobias,
+                               initialW=LeCunUniform(1 / numpy.sqrt(3)))
+
+            self.sigma = L.Linear(in_size, self.out_size, self.nobias,
+                                  initialW=VarianceScalingConstant(
+                                      sigma_scale),
+                                  initial_bias=VarianceScalingConstant(sigma_scale, fan='out'))
```
toslunar: Could you check the `flake8` error here? (too long line)

toslunar:
Now the expected variance in `test_noisy_linear.TestFactorizedNoisyLinear._test_randomness` should be computed by
```
target *= 2 / numpy.pi + numpy.sqrt(2 / numpy.pi) / fan_out
```
if the link has bias.

PR #290, opened 2018-08-01
Fixing pypi badge

jleni:


muupan:
Can you explain what does this PR try to fix? The current pypi badge seems to have no problem.

jleni:
it was not showing in my case. Likely it was a problem with shields.io

muupan:
Thanks anyway.

PR #149, opened 2017-10-11
A2C

iory:
Add A2C proposed by following thesis.
https://arxiv.org/abs/1708.05144
https://blog.openai.com/baselines-acktr-a2c/
A2C is a synchronous, deterministic variant of Asynchronous Advantage  Actor Critic (A3C).

PR #295, opened 2018-08-09
Batch PPO Implementation

ljvmiranda921:
Refactor branch: https://github.com/ljvmiranda921/chainerrl/pull/13

# Description

This PR is built on top of @iory 's A2C implementation in #149. It provides a batch/parallel implementation of [Proximal Policy Optimization](https://arxiv.org/abs/1707.06347). I'm using #149's `VecEnv` environment to achieve this task. Here are the main changes:

- Add batch implementation to PPO via the `batch_act()` and `batch_observe()` methods
- Add `chainerrl.experiments.train_agent_batch()` *Status: In progress*
- Add new tests for batch PPO *Status: In progress*

## Changes in data structure

Previously, the computation resides in the `self.memory` and `self.last_episode` attributes of PPO. Now we're also using `self.batch_memory` to handle this task. During a batch run, the type signature looks like:

```python
# Type signature for self.batch_memory during batch run
batch_memory :: [[dict], [dict], [dict]]
where len(batch_memory) == batch_env.num_envs
```

The same goes for `self.last_episode`:

```python
# Type signature for self.last_episode
last_episode :: [[dict], [dict], [dict]]
where len(last_episode) == batch_env.num_envs
```

## New methods: `batch_act()` and `batch_observe()`

These methods are supposed to handle all batch computations during the run. The method `batch_act(batch_obs)` performs a set of actions given a set of observations, while `batch_observe(obs, r, done, info)` updates the model. A simple way to use them can be seen below:

```python
t = 0; steps = 100
# o_0, r_0 : Init observation and reward
obs = batch_env.reset()
r = np.zeros(num_process, dtype='f')
# Initialize episode reward
episode_r = np.zeros(num_process, dtype='f')

while t < steps:
    # a_t : First action
    action = agent.batch_act(obs)
    # o_{t+1}, r_{t+1} : Get observation and reward
    obs, r, done, info = batch_env.step(action)
    # Train model
    agent.batch_observe(obs, r, done, info)
    # Update counters
    t += 1
    update_or_reset_reward(episode_r, done, info)
```

This assumes that the environment in `VectorEnv` sends a `reset` signal in the form of a dictionary entry in `info`.

halfcheetah-v2-5-finish
hopper-v2-5-finish
reacher-v2-5-finish
cartpole-v1
halfcheetah-v2
hopper-v2
pendulum-v0
reacher-v2


ljvmiranda921:
TODO: Merge `batch_memory` branch here

chainerrl/envs/vec_env.py
```diff
@@ -0,0 +1,106 @@
...
+        results = [remote.recv() for remote in self.remotes]
+        obs, rews, dones, infos = zip(*results)
+        return np.stack(obs), np.stack(rews), np.stack(dones), infos
+
+    def reset(self):
+        for remote in self.remotes:
+            remote.send(('reset', None))
+        return np.stack([remote.recv() for remote in self.remotes])
+
+    def close(self):
```
muupan: If `close` is not called, the program cannot exit due to remaining processes. This is cumbersome. I think we should do the same procedure in `VectorEnv.__del__` instead so that users do not need to care about it.

examples/gym/train_ppo_batch_gym.py
```diff
@@ -0,0 +1,240 @@
...
+    # Set a random seed used in ChainerRL
+    misc.set_random_seed(args.seed, gpus=(args.gpu,))
+
+    args.outdir = experiments.prepare_output_dir(args, args.outdir)
+
+    def make_env(test):
+        env = gym.make(args.env)
+        # Use different random seeds for train and test envs
+        env_seed = 2 ** 32 - 1 - args.seed if test else args.seed
+        env.seed(env_seed)
```
muupan: Each env in VectorEnv should be assigned different random seeds. See `train_a3c_gym.py` for how to assign different random seeds.

chainerrl/experiments/train_agent_batch.py
```diff
@@ -0,0 +1,222 @@
...
+            episode_r += r
+            episode_len += 1
+            episode_idx += 1 - masks
+            # Add to deque whenever done/reset
+            episode_r_ = np.ma.masked_array(episode_r, masks)
+            d.extend(episode_r_.compressed())
+            # Start new episode for those with mask
+            episode_r *= masks
+            episode_len *= masks
+            t += 1
```
muupan: `t` should be the total number of transitions experienced so far. So, it should increase by `num_envs`, not by 1. By doing so, we can keep the other hyperparameters and only change `num_envs` to trade cpus with computation time.

Issue #281, opened 2018-06-19
Abstract method is not overridden in some of the gaussian policies


uidilr:
I made PR #280 that update Policy class to inherit ABCMeta.
From Travis-CI log, it turned out that two gaussian policy classes,  https://github.com/chainer/chainerrl/blob/2be8d0b64140edffb0c9ac32b78f883786299f10/chainerrl/policies/gaussian_policy.py#L132 
and 
https://github.com/chainer/chainerrl/blob/2be8d0b64140edffb0c9ac32b78f883786299f10/chainerrl/policies/gaussian_policy.py#L203
does not override abstract method of gaussian policy. 

https://github.com/chainer/chainerrl/blob/2be8d0b64140edffb0c9ac32b78f883786299f10/chainerrl/policies/gaussian_policy.py#L27-L37  

Does function `compute_mean_and_var` has to be the abstract method?

uidilr:
I thought about two candidates of solutions to this issue.

####  1.  Implement method `compute_mean_and_var` for these two classes.
- Advantage
  - It follows the principle of objected-oriented.
- Disadvantage
  - It is redundant to implement unused method.

#### 2.  Inherit `Policy` class instead of `GaussianPolicy` 
- Advantage
  -  Less modification
- Disadvantage 
  - It is inconsistent to inherit `Policy` instead of `GaussianPolicy`, when the sub-class's name includes "GaussianPolicy". It is confusing and someone may try to call `compute_mean_and_var` of these sub-class and get error. 

I think that the first one is better.   
What do you think?

muupan:
Sorry for my late response. This needs to be fixed.

I'm for removing `GaussianPolicy` interface completely and inheriting `Policy` instead because `GaussianPolicy` does not help simplify implementing a Gaussian policy.

toslunar:
Closed by #305.

PR #305, opened 2018-09-04
Remove GaussianPolicy and obsolete policies

### LINKED ISSUE #276: Remove LinearGaussianPolicyWithSphericalCovariance and LinearGaussianPolicyWithDiagonalCovariance
since they can be replaced with FCGaussianPolicy, which is much more flexible.

muupan:
Resolves #276 and #281 

Since `LinearGaussianPolicyWithDiagonalCovariance` hard-codes squashing actions by tanh, replacing it with `FCGaussianPolicy` will change the behavior. I confirmed that `train_a3c_gym.py` can solve `InvertedPendulum-v1` as before by running `python train_a3c_gym.py 4 --env InvertedPendulum-v1 --arch LSTMGaussian --t-max 50`.


prabhatnagarajan:
LGTM;

PR #309, opened 2018-09-12
Make random access queue sampling code cleaner

prabhatnagarajan:
We have previously overridden the __getitem__ operator. However, the sampling method doesn't make use of this override, and instead replicates much of the code in that method. This commit does change the semantics of what occurs in practice. However, due to the uniformity of sampling, in expectation it does the same thing. In a nutshell, both the previous code as well as this commit implement uniform sampling. However, the indexes that are chosen are different between the previous code and this one.


toslunar:
It's OK because 40 us per iteration is not big.

PR #314, opened 2018-09-28
Fix the error caused by inexact delta_z

muupan:
- [x] check if the performance is affected


muupan:
Good news. This fix actually seems to boost performance.

image
image
image

PR #302, opened 2018-08-31
Tuned DoubleDQN with prioritized experience replay

muupan:
Merge #301 first.

This PR improves train_dqn_ale.py:
- use tuned DoubleDQN as better default settings (used in the Double DQN paper)
- add `--prioritized` to enable prioritized experience replay
- remove `--activation` and `--use-sdl`

PR #317, opened 2018-10-01
Adds N-step learning for DQN-based agents.

prabhatnagarajan:
Known affected agents:
-DQN
- Double DQN
- PAL
- Advantage Learning
- DPP
- SARSA
- Categorical DQN (TODO), waiting on bug fix


prabhatnagarajan:
Results:
Performs equal or slightly better in two domains: Qbert and Pong. Otherwise, N-step DQN performs worse.

Space Invaders
nstep

Breakout
breakout-nstep

Seaquest
seaquest-nstep

Qbert
qbert-nstep

Asterix
asterix-nstep

Pong
pong-nstep

BeamRider
beamrider-nstep

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Just eyeballed this PR's DQN (which is actually double DQN) results and compared them to the Tuned DQN results from:
https://github.com/chainer/chainerrl/pull/302#issuecomment-422310777.

The results look similar, so it suggests that the n-step DQN implementation of 1-step DQN does not adversely affect performance.

prabhatnagarajan:
Other affected agents:
- DDPG
- PCL

chainerrl/agents/dqn.py
```diff
@@ -201,28 +201,27 @@ def update(self, experiences, errors_out=None):
 
         This function is thread-safe.
         Args:
-          experiences (list): list of dict that contains
+          experiences (list): list of dicts that contains
```
muupan: It is `list of lists of dicts` now?

prabhatnagarajan: Yes, that's right. I have changes to make to the docstrings.

chainerrl/agents/dqn.py
```diff
@@ -318,22 +321,22 @@ def _compute_y_and_t(self, exp_batch, gamma):
...
 
-    def _compute_loss(self, exp_batch, gamma, errors_out=None):
+    def _compute_loss(self, exp_batch, errors_out=None):
         """Compute the Q-learning loss for a batch of experiences
 
 
         Args:
           experiences (list): see update()'s docstring
-          gamma (float): discount factor
+          discount (float): Amount by the Q-values should be discounted
```
muupan: This argument was removed.

chainerrl/replay_buffer.py
```diff
@@ -130,19 +131,28 @@ def stop_current_episode(self):
...
 
     def append(self, state, action, reward, next_state=None, next_action=None,
                is_state_terminal=False):
         experience = dict(state=state, action=action, reward=reward,
                           next_state=next_state, next_action=next_action,
                           is_state_terminal=is_state_terminal)
-        self.memory.append(experience)
+        self.last_n_transitions.append(experience)
+        if is_state_terminal:
+            self.memory.append(copy.copy(self.last_n_transitions))
```
muupan: How about `self.memory.append(list(self.last_n_transitions))` instead of copying a deque? `self.last_n_transitions` is a deque, but transitions appended to `self.memory` do not need to be a deque.

chainerrl/replay_buffer.py
```diff
@@ -160,7 +170,9 @@ def load(self, filename):
                 self.memory, maxlen=self.memory.maxlen)
 
     def stop_current_episode(self):
-        pass
+        if 0 < len(self.last_n_transitions) < self.num_steps:
```
muupan: Isn't is possible that `len(self.last_n_transitions) == self.num_steps`?

prabhatnagarajan: Yes, but then `self.last_n_transitions` would have been added to the memory already, and we wouldn't want to do it twice. This catches the edge case that the episode ends in less than `self.num_steps`, and nothing has been added to the memory. The important part of this method is that `self.last_n_transitions` is cleared, so that we don't merge transitions from the next episode to this one.

muupan: Ah I see. So `len(self.last_n_transitions) < self.num_steps` is correct.

muupan:
Can you resolve the conflicts?

chainerrl/agents/dqn.py
```diff
@@ -266,16 +268,18 @@ def update_from_episodes(self, episodes, errors_out=None):
                     transitions.append(ep[i])
```
muupan:
```suggestion
                     transitions.append([ep[i]])
```

chainerrl/agents/dqn.py
```diff
@@ -266,16 +268,18 @@ def update_from_episodes(self, episodes, errors_out=None):
...
                     if has_weights:
                         weights_step.append(weights[index])
-                batch = batch_experiences(transitions,
-                                          xp=self.xp,
-                                          phi=self.phi,
-                                          batch_states=self.batch_states)
+                batch = batch_experiences(
+                    [transitions],
```
muupan:
```suggestion
                    transitions,
```

chainerrl/agents/dqn.py
```diff
@@ -266,16 +268,18 @@ def update_from_episodes(self, episodes, errors_out=None):
...
+                    xp=self.xp,
+                    phi=self.phi,
+                    gamma=self.gamma,
+                    batch_states=self.batch_states)
```
muupan:
```suggestion
                    batch_states=self.batch_states)
                assert len(batch['state']) == len(transitions)
```

muupan: This is how `update_from_episodes` works:
- sample N episodes from `EpisodicReplayBuffer`
- take the first transition of each episode, batch those transitions, and compute a loss on that batch
- take the second transition of each episode, ...
- ...
- take the last transition of each episode, ...
- sum the losses, backprop, and update the model

Thus, `len(batch['state'])` should be equal to the number of transitions taken at each step.

chainerrl/agents/dqn.py
muupan: `epi[i]` is a dict that represents a transition.

tests/test_replay_buffer.py
```diff
@@ -18,52 +21,145 @@
...
+        four_step_transition.append(dict(
+                                    state=1, action=1, reward=1, next_state=1,
+                                    next_action=1, is_state_terminal=True))
+        experiences.append(four_step_transition)
+        batch = replay_buffer.batch_experiences(
+            experiences, np, lambda x: x, 0.99)
+        self.assertEqual(batch['state'][0], 1)
+        self.assertSequenceEqual(list(batch['is_state_terminal']),
+                                 list(np.asarray([0.0, 0.0, 1.0],
+                                                 dtype=np.float32)))
```
muupan: Can you test `batch['next_state']` and `batch['discount']` as well?

tests/test_replay_buffer.py
```diff
@@ -18,52 +21,145 @@
...
+                # [[0,0,0], [0, 0, 1], [0, 1], [1]]
+                if len(item) < self.num_steps:
+                    self.assertEqual(item[len(item) - 1]['state'], 1)
+                    for i in range(len(item) - 1):
+                        self.assertEqual(item[i]['state'], 0)
+                else:
+                    for i in range(len(item) - 1):
+                        self.assertEqual(item[i]['state'], 0)
+
+    def test_batch_experiences(self):
```
muupan: Since `batch_experiences` is a function independent of `ReplayBuffer`, I think it is better to create a separate TestCase class for it.

prabhatnagarajan:
The previous version of N-step DQN did not add an additional N-1 transitions when an episode finishes. These are the updated results after adding this feature to N-step learning.
asterix-nstep
beamrider-nstep
breakout-nstep
pong-nstep
qbert-nstep
seaquest-nstep
spaceinvaders-nstep

muupan:
Looks good!

PR #324, opened 2018-10-04
Support gym==0.10.8

### LINKED ISSUE #323: Support gym==0.10.8
- gym.undo_logger_setup was removed 
- we need to explicitly import `gym.spaces` and `gym.wrappers` now

https://github.com/openai/gym/commit/e689f93a425d97489e590bba0a7d4518de0dcc03#diff-9c9bed4efd3db64376fabec6bfd32891

muupan:
This PR makes following changes to support gym==0.10.8.
- remove calls of gym.undo_logger_setup
- remove chainerrl/spaces.py
- import gym.spaces explicitly
- import gym.wrappers explicitly

Support of gym older than 0.9.7 is dropped.

Resolves #323, #325.


muupan:
CI was broken by the update of jupyter-console but it's also fixed by this PR.

muupan:
I confirmed that all the tests passed.

chainerrl/envs/abc.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 
 from chainerrl import env
-from chainerrl import spaces
+from gym import spaces
```
toslunar: Could you fix the grouping of the imports?

chainerrl/envs/ale.py
```diff
@@ -13,7 +13,7 @@
 import numpy as np
 
 from chainerrl import env
-from chainerrl import spaces
+from gym import spaces
```
toslunar: Could you fix the grouping of the imports?

muupan:
I fixed the grouping of imports.

PR #332, opened 2018-10-16
Stop caching the result of numpy.concatenate in LazyFrames

muupan:
This change saves a lot of memory. See https://github.com/openai/baselines/issues/295.

- [x] check if it does not affect scores


muupan:
I compared the results of `examples/ale/train_dqn_ale.py --eval-interval 1000000 --prioritized --lr 6.25e-5 --env BreakoutNoFrameskip-v4` before and after this PR. It seems this PR does not affect the scores.

image

PR #333, opened 2018-10-16
Merge consecutive with statements

muupan:
for less indent.


muupan:
I confirmed all the tests passed.

PR #336, opened 2018-10-17
Use Variable.array instead of Variable.data

### LINKED ISSUE #335: Use Variable.array instead of Variable.data
Since Chainer v3, `chainer.Variable` has the `array` property that is equivalent to the `data` property. It is better to use `array` because it can catch a bug of unintentionally accessing `numpy.ndarray.data`.

https://docs.chainer.org/en/stable/reference/generated/chainer.Variable.html#chainer.Variable.data

> Note that using this attribute directly is discouraged; use array instead. Using array, you can find an error earlier when your code mixes up Variable and ndarray because ndarray does not have an attribute .array while it has .data.

muupan:
Resolves #335 

Chainer v2 support is dropped.

### DIFF
--- a/chainerrl/functions/sum_arrays.py
+++ b/chainerrl/functions/sum_arrays.py
@@ -21,7 +21,7 @@ class SumArrays(function.Function):
 
     def forward_gpu(self, inputs):
         n = len(inputs)
-        ptrs = cuda.cupy.asarray([x.data.ptr for x in inputs],
+        ptrs = cuda.cupy.asarray([x.array.ptr for x in inputs],
                                  dtype=cuda.cupy.int64)
         y = cuda.elementwise(
             'T x0, int64 xs, int32 n_xs',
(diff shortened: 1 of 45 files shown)


toslunar:
Once we drop Chainer v2 support, we can always use `F.matmul` in place of `matmul_v3`.  This can be done in another PR.

muupan:
It seems pytest-cov was missing for v3.1.0 CI, so I added.

chainerrl/misc/prioritized.py
```diff
@@ -17,15 +17,15 @@ class PrioritizedBuffer (object):
     def __init__(self, capacity=None, wait_priority_after_sampling=True,
                  initial_max_priority=1.0):
         self.capacity = capacity
-        self.data = collections.deque()
+        self.array = collections.deque()
```
toslunar: `.data` in the file is not an attribute of `Variable`.

chainerrl/functions/sum_arrays.py
```diff
@@ -21,7 +21,7 @@ def backward(self, inputs, grads):
 
     def forward_gpu(self, inputs):
         n = len(inputs)
-        ptrs = cuda.cupy.asarray([x.data.ptr for x in inputs],
+        ptrs = cuda.cupy.asarray([x.array.ptr for x in inputs],
```
toslunar: Please revert the changes to the file.

muupan:
Thanks for your great review!

PR #337, opened 2018-10-17
Remove code for Chainer v2

muupan:
This PR removes code that was needed to support Chainer v2.


muupan:
I confirmed that all the tests pass.

PR #343, opened 2018-10-31
Close envs in async training

### LINKED ISSUE #342: How to close environments in experiments.train_agent_async
My environment needs to be closed after its use, but I couldn't find the way to close environments distributed in `experiments.train_agent_async` after training is finished.

muupan:
- Resolves #342 
- Corrects the termination condition of `train_agent_async`.
- Adds tests for `train_agent_async`

I confirmed that all the tests pass.

### DIFF
--- a/chainerrl/experiments/train_agent_async.py
+++ b/chainerrl/experiments/train_agent_async.py
@@ -236,6 +236,8 @@ def train_agent_async(outdir, processes, make_env,
             f()
 
         env.close()
+        if eval_env is not env:
+            eval_env.close()
 
     async_.run_async(processes, run_func)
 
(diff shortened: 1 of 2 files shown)


muupan:
It seems that `eval_env` need to be closed, too.

muupan:
I confirmed again that all the tests pass.

chainerrl/experiments/train_agent_async.py
```diff
@@ -87,7 +87,7 @@ def train_loop(process_idx, env, agent, steps, outdir, counter,
                 for hook in global_step_hooks:
                     hook(env, agent, global_t)
 
-                if global_t > steps or training_done.value:
+                if global_t >= steps or training_done.value:
```
toslunar: Could you also fix the line 101?

muupan: Fixed and added a test of it.

PR #353, opened 2018-11-14
Makes DQN example match Deepmind

prabhatnagarajan:


prabhatnagarajan:
Evaluation protocols todos:

- 30 evaluation episodes of best network

prabhatnagarajan:
Implementation todos:

- Episodes are reset on the loss of a life during training (**Source**: https://github.com/deepmind/dqn/blob/master/dqn/train_agent.lua#L87 passes in true, https://github.com/deepmind/alewrap/blob/master/alewrap/GameEnvironment.lua#L102 returns true if the life is lost and true is passed to the function, and https://github.com/deepmind/dqn/blob/master/dqn/train_agent.lua#L90 and https://github.com/deepmind/alewrap/blob/master/alewrap/GameEnvironment.lua#L131 starts a new game)

prabhatnagarajan:
Sample output of the best scores:

```
The results of the best scoring network:

median:1.0
max:4.0
min:0.0
stdev:1.1426929274467317
mean:1.7333333333333334
episodes:30
```

prabhatnagarajan:
stargunner-dqn
qbert-dqn
pinball-dqn
asterix-dqn

muupan:
What happened in Pinball?

prabhatnagarajan:
A fresh set of scores.

asterix-dqn
beamrid-dqn
breakout-dqn
pong-dqn
qbert-dqn
seaquest-dqn
spaceinvad-dqn

prabhatnagarajan:
The scores are (for a single run)
Asterix: 5755.0 (us) vs. **6012** (deepmind)
BeamRider: **6942** (us) vs. 6846 (deepmind)
Breakout: 382.7 (us) vs. **401.2** (deepmind)
Seaquest: **6364** (us) vs. 5286 (deepmind)
Pong: **20.133** (us) vs. 18.9 (deepmind)
SpaceInvaders: 1555.67 (us) vs. **1976** (deepmind)
Qbert: 7482.5 (us) vs. **10596** (deepmind)

chainerrl/wrappers/continuing_time_limit.py
```diff
@@ -38,8 +38,9 @@ def step(self, action):
         observation, reward, done, info = self.env.step(action)
         self._elapsed_steps += 1
 
-        if self._max_episode_steps <= self._elapsed_steps:
-            info['needs_reset'] = True
+        if self._max_episode_steps:
+            if self._max_episode_steps <= self._elapsed_steps:
+                info['needs_reset'] = True
```
muupan: If `max_episode_steps` is None, it does not make sense to wrap env with `ContinuingTimeLimit`. I think it is better to modify `make_atari` so that it skips `ContinuingTimeLimit` if `max_frames` is None, not modifying `ContinuingTimeLimit`.

chainerrl/experiments/train_agent.py
```diff
@@ -88,6 +88,9 @@ def train_agent(agent, env, steps, outdir, max_episode_len=None,
 
     # Save the final model
     save_agent(agent, t, outdir, logger, suffix='_finish')
+    if evaluator:
+        if evaluator.save_best_so_far_agent and not evaluator.saved:
+            save_agent(agent, "best", outdir, logger, suffix='')
```
muupan: Saving the agent as "best" when no evaluation is performed is confusing. I think it is better not to save at all.

examples/atari/dqn/train_dqn.py
```diff
@@ -169,22 +135,45 @@ def phi(x):
...
+            save_best_so_far_agent=True,
             eval_env=eval_env,
         )
 
+        dir_of_best_network = os.path.join(args.outdir, "best")
+        agent.load(dir_of_best_network)
+
+        # Change seed for final evaluation
+        final_seed = max(train_seed, test_seed) - \
+            min(train_seed, test_seed) - 1
```
muupan: Can you elaborate this calculation of `final_seed`?

examples/atari/dqn/train_dqn.py
```diff
@@ -169,22 +135,45 @@ def phi(x):
...
+            agent=agent,
+            n_steps=None,
+            n_episodes=args.n_best_episodes,
+            max_episode_len=4500,
+            logger=None)
+        stats["episodes"] = args.n_best_episodes
+        with open(os.path.join(args.outdir, 'bestscores.txt'), 'a+') as f:
+            print("The results of the best scoring network:\n", file=f)
+            for stat in stats:
+                print(str(stat) + ":" + str(stats[stat]), file=f)
```
muupan: I think it is better to use an easy-to-parse format like json to save `stats`. It would be still helpful to print a human-readable text like this to stdout though.

examples/atari/dqn/train_dqn.py
prabhatnagarajan: I'm trying to ensure that the train, test, and the final seed are different. We know that train and test seeds are different. I'm trying to get a third number that's different from the train and test seed. This is supposed to do that... Not sure it's the best way though. Ideally we do something like max(seeds) - min(seeds) to get a unique nonnegative seed.  However, if one of the seeds is 0, then max(seeds) - min(seeds) = max(seeds), and we'll have final_seed= max_seed. This ensures that doesn't happen.

muupan: The algorithm does not work if seeds=[1,3], and it can produce a negative value.

If we use `eval_env` for both regular and final evaluation, I think it is ok to use `eval_env` for final evaluation without without additionally calling `eval_env.seed(final_seed)`, since its internal random state is already changed.

examples/atari/dqn/train_dqn.py
```diff
@@ -167,7 +167,7 @@ def phi(x):
             logger=None)
         with open(os.path.join(args.outdir, 'bestscores.txt'), 'w') as f:
```
muupan: `bestscores.json` is better since it's json.

examples/atari/dqn/train_dqn.py
```diff
@@ -169,22 +136,42 @@ def phi(x):
...
+        stats = experiments.evaluator.eval_performance(
+            env=eval_env,
+            agent=agent,
+            n_steps=None,
+            n_episodes=args.n_best_episodes,
+            max_episode_len=4500,
+            logger=None)
+        with open(os.path.join(args.outdir, 'bestscores.txt'), 'w') as f:
+            json_stats = json.dumps(stats)
+            print(str(json_stats), file=f)
```
muupan: You can just write `json.dump(stats, f)`

chainerrl/experiments/evaluator.py
```diff
@@ -312,6 +313,7 @@ def __init__(self,
         self.prev_eval_t = (self.step_offset -
                             self.step_offset % self.eval_interval)
         self.save_best_so_far_agent = save_best_so_far_agent
+        self.saved = False
```
muupan: `self.saved` seems no longer used anywhere.

chainerrl/experiments/evaluator.py
```diff
@@ -342,7 +344,8 @@ def evaluate_and_update_max_score(self, t, episodes):
                              self.max_score, mean)
             self.max_score = mean
             if self.save_best_so_far_agent:
-                save_agent(self.agent, t, self.outdir, self.logger)
+                save_agent(self.agent, "best", self.outdir, self.logger)
+                self.saved = True
```
muupan: Ditto.

PR #356, opened 2018-11-16
Allow envs to send a 'needs_reset' signal

muupan:
This PR enables an env to send a signal that indicates it needs a reset via the `info` dict returned by `env.step` without setting `done=True`.

This functionality is needed to implement to strictly follow the training protocol of DeepMind on Atari games, which limits the number of frames of an episode ignoring life losses, while the agent still sees episodes that terminate by a life loss.

- [x] add ContinuingTimeLimit wrapper
- [x] handle needs_reset signal in both training and evaluation
- [x] check how it affects DQN on Atari

### DIFF
--- a/chainerrl/experiments/train_agent.py
+++ b/chainerrl/experiments/train_agent.py
@@ -59,10 +59,9 @@ def train_agent(agent, env, steps, outdir, max_episode_len=None,
             for hook in step_hooks:
                 hook(env, agent, t)
 
-            if (done
-                    or episode_len == max_episode_len
-                    or t == steps
-                    or info.get('needs_reset', False)):
+            reset = (episode_len == max_episode_len
+                     or info.get('needs_reset', False))
+            if done or reset or t == steps:
                 agent.stop_episode_and_train(obs, r, done=done)
                 logger.info('outdir:%s step:%s episode:%s R:%s',
                             outdir, t, episode_idx, episode_r)
--- a/chainerrl/experiments/train_agent.py
+++ b/chainerrl/experiments/train_agent.py
@@ -59,7 +59,10 @@ def train_agent(agent, env, steps, outdir, max_episode_len=None,
             for hook in step_hooks:
                 hook(env, agent, t)
 
-            if done or episode_len == max_episode_len or t == steps:
+            if (done
+                    or episode_len == max_episode_len
+                    or t == steps
+                    or info.get('needs_reset', False)):
                 agent.stop_episode_and_train(obs, r, done=done)
                 logger.info('outdir:%s step:%s episode:%s R:%s',
                             outdir, t, episode_idx, episode_r)
--- a/chainerrl/experiments/train_agent_async.py
+++ b/chainerrl/experiments/train_agent_async.py
@@ -28,7 +28,6 @@ def train_loop(process_idx, env, agent, steps, outdir, counter,
 
     try:
 
-        total_r = 0
         episode_r = 0
         global_t = 0
         local_t = 0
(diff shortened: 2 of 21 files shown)


muupan:
I confirmed the change introduced by this PR does not affect the scores on Atari.

old: `examples/ale/train_dqn_ale.py --eval-interval 1000000 --env {env_id}` (before this PR)
new: `examples/ale/train_dqn_ale.py --eval-interval 1000000 --max-frames 108000 --env {env_id}` (after this PR)

beamridernoframeskip-v4
breakoutnoframeskip-v4
seaquestnoframeskip-v4

chainerrl/experiments/train_agent.py
```diff
@@ -59,7 +59,9 @@ def train_agent(agent, env, steps, outdir, max_episode_len=None,
             for hook in step_hooks:
                 hook(env, agent, t)
 
-            if done or episode_len == max_episode_len or t == steps:
+            reset = (episode_len == max_episode_len
+                     or info.get('needs_reset', False))
+            if done or reset or t == steps:
```
prabhatnagarajan: Later in this "if" statement (https://github.com/chainer/chainerrl/pull/356/files#diff-a2caf3ec0e2750a1d16edb375789daa5R81), you reset the environment. Why do you reset the environment if done is true? What if reset = False?

chainerrl/experiments/train_agent_async.py
```diff
@@ -41,19 +40,34 @@ def train_loop(process_idx, env, agent, steps, outdir, counter,
...
+            if done or reset or global_t >= steps or training_done.value:
                 agent.stop_episode_and_train(obs, r, done)
+
                 if process_idx == 0:
                     logger.info(
                         'outdir:%s global_step:%s local_step:%s R:%s',
                         outdir, global_t, local_t, episode_r)
                     logger.info('statistics:%s', agent.get_statistics())
+
+                # Evalaute the current agent
```
prabhatnagarajan: Nit: Evaluate*, not "Evalaute"

chainerrl/experiments/train_agent_batch.py
```diff
@@ -86,9 +88,6 @@ def train_agent_batch(agent, env, steps, outdir, log_interval=None,
             #   5. reset the env to start a new episode
```
prabhatnagarajan: Should these comments be revised?

chainerrl/wrappers/atari_wrappers.py
```diff
@@ -10,6 +10,8 @@
 
 from gym import spaces
 
+import chainerrl
```
prabhatnagarajan: Since you're changing the file, perhaps we should change the header from "This file is a fork from a MIT-licensed project" to "This file _adapted_ from an MIT-licensed project..."

chainerrl/wrappers/continuing_time_limit.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class ContinuingTimeLimit(gym.Wrapper):
+    """TimeLimit wrapper for continuing environments.
+
+    This is similar gym.wrappers.TimeLimit, which sets a time limit for
+    each episode, except that done=False is returned and that
+    info['needs_reset'] is set to True when past the limit.
+
+    Code that calls env.step is repsonsible for checking the info dict, the
```
prabhatnagarajan: "responsible" - typo

chainerrl/experiments/train_agent.py
muupan: Even if `reset == False`, we need to call env.reset() if `done == True`. In other words, the `reset` variable can be false if we reset the env due to `done == True`. It is possible to rename `reset` as `non_done_reset` or something, but it would be verbose.

prabhatnagarajan: So this is targeted primarily at environments that reset based off of a `max-episode-length` or via `done`, and not environments where `done` is `True` but you still do not reset the environment?

muupan: Correct. Currently, ChainerRL assumes that `env.reset` must be called when `done==True`.

examples/ale/train_dqn_ale.py
```diff
@@ -103,9 +103,9 @@ def main():
                         help='Network architecture to use.')
     parser.add_argument('--steps', type=int, default=5 * 10 ** 7,
                         help='Total number of timesteps to train the agent.')
-    parser.add_argument('--max-episode-len', type=int,
-                        default=30 * 60 * 60 // 4,  # 30 minutes with 60/4 fps
-                        help='Maximum number of timesteps for each episode.')
+    parser.add_argument('--max-frames', type=int,
```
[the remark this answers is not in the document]
muupan: Other examples that use atari_wrappers are affected as well. If the changes to examples/ale/train_dqn_ale.py look ok to you, I can apply the same changes to other examples as well, though I believe they would work without changes.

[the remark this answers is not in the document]
prabhatnagarajan: Sounds good. I think we should still make the changes to other examples using atari_wrappers for completeness/consistency.

chainerrl/experiments/train_agent_batch.py
muupan: I think these comments are still correct except that 3-5 are skipped when training is finished. I'll clarify this in the comments.

chainerrl/wrappers/atari_wrappers.py
muupan: "fork" means it already has changes, no?

muupan:
I addressed the comments and made the same change to other examples.

PR #367, opened 2018-12-10
Implements Time-based evaluations

prabhatnagarajan:
This PR enables time-based evaluations in order to mirror deepmind's evaluation protocols more accurately. Previous versions of ChainerRL only supported episodic evaluations (i.e. evaluations spanning a certain number of episodes). However, **this PR does not support Asynchronous time-based evaluations**.


tests/experiments_tests/test_evaluator.py
```diff
@@ -99,52 +118,100 @@ def test_evaluate_if_necessary(self):
...
         else:
             self.assertEqual(agent.save.call_count, 0)
 
 
+@testing.parameterize(
+    *testing.product({
+        'n_episodes': [None, 1],
+        'n_timesteps': [2, 5, 6],
+    })
+)
```
muupan: `test_needs_reset` does not depend on these parameters, but now it is run 2*3 times because of them. Can you split the class and add `@testing.parameterize` to `test_timesteps` only?

tests/experiments_tests/test_evaluator.py
```diff
@@ -159,15 +226,83 @@ def test_needs_reset(self):
...
         self.assertAlmostEqual(scores[0], 0)
         self.assertAlmostEqual(scores[1], 0.5)
 
 
+@testing.parameterize(
+    *testing.product({
+        'n_episodes': [None, 1],
+        'n_timesteps': [2, 5, 6],
+    })
+)
```
muupan: Ditto.

chainerrl/experiments/evaluator.py
```diff
@@ -122,39 +152,68 @@ def batch_run_evaluation_episodes(
...
+                and completed_episode == 0
+                and episode_len[0] >= n_steps):
+            eval_episode_returns = [episode_r[0]]
+            eval_episode_lens = [n_steps]
+            termination_conditions = True
+
+        if termination_conditions:
+            break
+        else:
+            obss = env.reset(not_end)
```
muupan: The logic of making `eval_episode_returns` and `eval_episode_lens` looks wrong to me. To be unbiased, we need to use episodes with smaller `episode_idx` first, even when episodes with larger `episode_idx` finish earlier. That is,

- if n_steps is not None
  - check episode_idx=0,1,... sequentially until you first encounter an unfinished episode, then sum up their lengths (including last unfinished one). if >=n_steps, then terminate.
- if n_episodes is not None
  - check if episode_idx=0,1,...,(n_episodes-1) are all finished. if yes, then terminate. episode_idx=n_episodes,... should not be used at all.

chainerrl/experiments/evaluator.py
```diff
@@ -212,19 +275,28 @@ class Evaluator(object):
...
                  save_best_so_far_agent=True,
                  logger=None,
                  ):
+        try:
+            assert (n_steps is None) != (n_episodes is None)
+        except AssertionError as error:
+            print("One of n_steps or n_episodes must be None. " +
+                  "Either we evaluate for a specified number " +
+                  "of episodes or for a specified number of timesteps.")
+            raise error
```
muupan: You can write as `assert condition, "message"`

chainerrl/experiments/evaluator.py
prabhatnagarajan: As discussed offline, the code is correct, though suboptimal. Will make the changes to improve code readability and efficiency.

chainerrl/experiments/evaluator.py
```diff
@@ -122,39 +152,81 @@ def batch_run_evaluation_episodes(
...
+        while first_unfinished_episode in episode_returns:
+            first_unfinished_episode += 1
+
+        # Check for termination conditions
+        eval_episode_returns = []
+        eval_episode_lens = []
+        if n_steps is not None:
+            total_time = 0
+            index = 0
+            while index < first_unfinished_episode:
```
muupan: `for index in range(first_unfinished_episode):` would be better

muupan:
It looks good except for suggested changes.

muupan:
Can you fix the autopep8 error of CI?

muupan:
I think it's ok to merge. CI failures should be fixed by #378

PR #378, opened 2019-01-08
Use AbnormalExitCodeWarning for nonzero exitcode warnings

muupan:
When subprocesses exit with nonzero exitcode, `run_async` raise warnings. Some tests check the number of warnings, but they do not work when there is another code that raises warnings, leading to failure in CI of current master.

This PR specifies a warning category for nonzero exitcode, `AbnormalExitCodeWarning`, and the tests will check warnings of that category.

### DIFF
--- a/chainerrl/misc/async_.py
+++ b/chainerrl/misc/async_.py
@@ -15,7 +15,7 @@ import numpy as np
 from chainerrl.misc import random_seed
 
 
-class AbnormalExitCodeWarning(Warning):
+class AbnormalExitWarning(Warning):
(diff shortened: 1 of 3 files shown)


chainerrl/misc/async_.py
```diff
@@ -15,6 +15,11 @@
 from chainerrl.misc import random_seed
 
 
+class AbnormalExitCodeWarning(Warning):
```
toslunar: How about `AbnormalExitWarning` or `AbnormalTerminationWarning`?  A nonzero `Process.exitcode` represents either an exitcode or a signal.

PR #390, opened 2019-01-28
Make NonbiasWeightDecay work again

### LINKED ISSUE #388: NonbiasWeightDecay does not contain a `timing` attribute
The hook function `NonbiasWeightDecay` does not contain a `timing` attribute, and hence the `add_hook` function raises a warning when using that hook. Can you please explain what is the more reasonable choice for `timing` in this case, when choosing between `pre` or `post`, and why?

muupan:
`NonbiasWeightDecay` does not work due to API changes of Chainer. This PR fixes this issue and adds tests for it.

Resolves #388


muupan:
I confirmed that both cpu and gpu tests passed.

PR #394, opened 2019-02-08
Fix DQN.update's docstring

muupan:
- it is no longer thread-safe
- its format was not consistent with other docstrings
- information is added

### DIFF
--- a/chainerrl/agents/dqn.py
+++ b/chainerrl/agents/dqn.py
@@ -213,7 +213,7 @@ class DQN(agent.AttributeSavingMixin, agent.BatchAgent):
                   - state (object): State
                   - action (object): Action
                   - reward (float): Reward
-                  - is_state_terminal (bool): True iff the state is terminal
+                  - is_state_terminal (bool): True iff next state is terminal
                   - next_state (object): Next state
                   - weight (float, optional): Weight coefficient. It can be
                     used for importance sampling.
--- a/chainerrl/agents/dqn.py
+++ b/chainerrl/agents/dqn.py
@@ -207,18 +207,21 @@ class DQN(agent.AttributeSavingMixin, agent.BatchAgent):
     def update(self, experiences, errors_out=None):
         """Update the model from experiences
 
-        This function is thread-safe.
         Args:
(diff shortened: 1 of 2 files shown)


chainerrl/agents/dqn.py
```diff
@@ -207,18 +207,21 @@ def call_test(self_, x):
...
+                  - is_state_terminal (bool): True iff the state is terminal
+                  - next_state (object): Next state
+                  - weight (float, optional): Weight coefficient. It can be
+                    used for importance sampling.
+            errors_out (list or None): If set to a list, then TD-errors
+                computed from the are appended to the list.
+
         Returns:
-          None
+            None
```
prabhatnagarajan: Not too familiar with the docstring formatting, this tab shouldn't matter right?

chainerrl/agents/dqn.py
```diff
@@ -207,18 +207,21 @@ def call_test(self_, x):
...
+                For DQN, each dict must contains:
+                  - state (object): State
+                  - action (object): Action
+                  - reward (float): Reward
+                  - is_state_terminal (bool): True iff the state is terminal
+                  - next_state (object): Next state
+                  - weight (float, optional): Weight coefficient. It can be
+                    used for importance sampling.
+            errors_out (list or None): If set to a list, then TD-errors
+                computed from the are appended to the list.
```
prabhatnagarajan: TD-error computed from the what?

chainerrl/agents/dqn.py
muupan: This is not a tab but two spaces. Google style docstring uses 4 spaces for an indentation. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

chainerrl/agents/dqn.py
muupan: sorry I added missing words

muupan:
```is_state_terminal (bool): True iff the state is terminal```
may be confusing because it is whether `next_state` is terminal, not whether `state` is terminal, so I modified it to
```is_state_terminal (bool): True iff next state is terminal```.

PR #405, opened 2019-02-26
Add tests of IQN examples

muupan:
Tests of IQN examples were missing, so this PR adds them to `test_examples.sh`. Updated the examples to make them work with latest API.


muupan:
I confirmed that both `./test_examples.sh -1` and `./test_examples.sh 0` work.

PR #410, opened 2019-03-10
Make copy_param support scalar parameters

muupan:
Current `copy_param` does not support scalar parameters due to use of `[:]`. This PR fixes the issue by switching to `[...]`.


toslunar:
I tried to restart the Travis build after #411 is merged, but it was not tested with Python 3.6.  Could you merge `master`?

muupan:
Now tested with Python 3.6.

PR #411, opened 2019-03-10
Use Python 3.6 in Travis CI

muupan:
This PR fixes the error of `from typing import Type` in Travis CI caused by `tornado==6.0.1` by switching from Python 3.5 to 3.6.

https://travis-ci.org/chainer/chainerrl/jobs/504242079


toslunar:
LGTM, too

PR #414, opened 2019-03-18
Benchmarks DQN example

prabhatnagarajan:
This PR includes the following:

- Benchmarking of the scores attained by the DQN example
- A comparison of the scores against the reported DQN results
- Information on the average training time of the DQN
- Information on the evaluation protocol used.

### DIFF
--- a/examples/atari/dqn/README.md
+++ b/examples/atari/dqn/README.md
@@ -28,10 +28,10 @@ The summary of the results is as follows:
  - These results are averaged over 5 runs per domain
  - We ran this example on 59 Atari domains. 
  - The original DQN paper paper ran results on 49 Atari domains. Within these 49 domains the results are as follows:
- 	- The reported results from the DQN paper are higher than ChainerRL on 26 domains.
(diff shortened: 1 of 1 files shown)


examples/atari/dqn/README.md
```diff
@@ -22,70 +22,95 @@ python train_dqn.py [options]
 To view the full list of options, either view the code or run the example with the `--help` option.
 
 ## Results
-These results reflect ChainerRL  `v0.5.0`.
```
muupan: What do you mean by `v0.5.0`/`v0.6.0`? I think it should be `v0.6.0` since `v0.5.0` does not have time-based evaluation.

prabhatnagarajan: You're correct, these results were obtained using ChainerRL `v0.5.0` as well as ChainerRL `v0.6.0`, but it seems that all the features I was using were integrated in ChainerRL `v0.6.0`. I will change it.

muupan:
The reported scores look correct.

PR #416, opened 2019-03-22
Enables batch DDPG agents to be trained.

prabhatnagarajan:


prabhatnagarajan:
This PR changes two Q-functions: `FCBNLateActionSAQFunction` and `FCLateActionSAQFunction`. A simple search in the ChainerRL repository:

https://github.com/chainer/chainerrl/search?q=FCBNLateActionSAQFunction&unscoped_q=FCBNLateActionSAQFunction

https://github.com/chainer/chainerrl/search?q=FCLateActionSAQFunction&unscoped_q=FCLateActionSAQFunction

shows that the DDPG example is the only place these are used. Furthermore, looking at the `__init__` functions for those networks, it appears that neither of them need recurrence. Since

https://github.com/chainer/chainerrl/blob/15d7cbbc8a1fe5b73d8c93cd3fc5cf2c1361f579/chainerrl/links/mlp.py#L16

and 

https://github.com/chainer/chainerrl/blob/15d7cbbc8a1fe5b73d8c93cd3fc5cf2c1361f579/chainerrl/links/mlp_bn.py#L31

are not recurrent.

prabhatnagarajan:
Swimmer-v2-BatchDDPGBN
Swimmer-v2-DDPGBN

chainerrl/agents/ddpg.py
```diff
@@ -344,6 +344,96 @@ def act(self, obs):
...
+        Returns:
+            Sequence of ~object: Actions.
+        """
+
+        batch_greedy_action = self.batch_act(batch_obs)
+        batch_action = [
+            self.explorer.select_action(
+                self.t, lambda: batch_greedy_action[i])
+            for i in range(len(batch_greedy_action))]
+        self.t += 1
```
muupan: This line should be removed since `self.t` is incremented in `batch_observe_and_train`.

chainerrl/agents/ddpg.py
```diff
@@ -344,6 +344,96 @@ def act(self, obs):
...
+
+        # Update the target network
+        if self.t % self.target_update_interval == 0:
+            self.sync_target_network()
```
muupan: These should be also removed since `batch_observe_and_train` has the same code.

PR #427, opened 2019-03-31
Add Branched and use it to simplify train_ppo_batch_gym.py

muupan:
~Merge #425 first.~

This PR adds a link named `ParallelLink` to just combine two independent callable links. This can be used to combine a policy and a value function instead of using `a3c.A3CModel`. I believe using `ParallelLink` whenever possible makes code more readable than designing agent-specific model interfaces.

To make it easier to write a Gaussian policy using `chainer.Sequential`, this PR also adds a link named `GaussianHeadWithStateIndependentCovariance`, which is just a head part of `FCGaussianPolicyWithStateIndependentCovariance`.

Todos:
- [x] Check training results before and after this PR


muupan:
The PR also slightly modified hyperparameters to make it more closer to openai/baselines' ppo2.

- Set `var_func=lambda x: F.exp(2 * x)` to parameterize log standard deviation. Before this PR `softplus` was used.
- Use `LeCunNormal(1e-2)` for both policy and value function. Before this PR, it was used only for policy.

Still, its weight initialization is different from ppo2, which uses orthogonal initialization. However, to use it, we need to wait https://github.com/chainer/chainer/pull/6031 to be merged into Chainer, so I leave it to another future PR.

image
image

muupan:
As for `Parallel` vs `ParallelLink`, maybe the former is better for consistency with Sequential. It is true the word "parallel" itself is confusing. How about "Branched"?

muupan:
After the offline discussion, I renamed `ParallelLink` to `Branched`.

examples/gym/train_ppo_batch_gym.py
```diff
@@ -164,14 +110,49 @@ def make_batch_env(test):
     obs_normalizer = chainerrl.links.EmpiricalNormalization(
         obs_space.low.size, clip_threshold=5)
 
+    winit_last = chainer.initializers.LeCunNormal(1e-2)
+
     # Switch policy types accordingly to action space types
-    if args.arch == 'FFSoftmax':
-        model = A3CFFSoftmax(obs_space.low.size, action_space.n)
-    elif args.arch == 'FFMellowmax':
-        model = A3CFFMellowmax(obs_space.low.size, action_space.n)
```
toslunar: I approve to remove mellowmax from the example.

muupan: Yes, I'm aware that this PR removes mellowmax model. I think nobody uses it.

PR #438, opened 2019-04-08
Increase tolerance of TestGaussianDistribution.test_entropy since sometimes it failed

muupan:
https://travis-ci.org/chainer/chainerrl/jobs/516785404


muupan:
Oops, since I restarted that job, I cannot see the failed log now...

PR #442, opened 2019-04-12
Pass env_id to replay buffer methods to correctly support batch training

muupan:
~Merge #443 before this PR.~

Current replay buffers with num_steps > 1 or episodic are not correct in batch training because they cannot know which env a given transition came from.

This PR adds the `env_id` argument to two methods of replay buffers: `append` and `stop_current_episode`. From `env_id` replay buffers can know which env a given transition came from and which env's episode is stopped.

TODO:
- [x] check how it affects scores with n-step return and batch training

### DIFF
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -156,7 +156,7 @@ class ReplayBuffer(AbstractReplayBuffer):
             next_state=next_state,
             next_action=next_action,
             is_state_terminal=is_state_terminal,
-            **kwargs,
+            **kwargs
         )
         last_n_transitions.append(experience)
         if is_state_terminal:
--- a/chainerrl/replay_buffer.py
+++ b/chainerrl/replay_buffer.py
@@ -29,7 +29,7 @@ class AbstractReplayBuffer(with_metaclass(ABCMeta, object)):
 
     @abstractmethod
     def append(self, state, action, reward, next_state=None, next_action=None,
-               is_state_terminal=False):
+               is_state_terminal=False, env_id=0, **kwargs):
         """Append a transition to this replay buffer.
 
         Args:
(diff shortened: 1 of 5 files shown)


muupan:
I checked the effect of the bug fixed by this PR. I added the `--n-step-return` option to `examples/ale/train_dqn_batch_ale.py` and run it with `--n-step-return 1` and `--n-step-return 3`, using 3 different random seeds:
```
python3 examples/ale/train_dqn_batch_ale.py --num-envs 8 --n-step-return 1 --steps 10000000 --env SpaceInvadersNoFrameskip-v4
```
```
python3 examples/ale/train_dqn_batch_ale.py --num-envs 8 --n-step-return 3 --steps 10000000 --env SpaceInvadersNoFrameskip-v4
```

Before this PR, `--n-step-return 3` completely failed. After this PR, `--n-step-return 3` learned faster than `--n-step-return 1` as expected.

image

chainerrl/replay_buffer.py
```diff
@@ -135,35 +143,44 @@ def __init__(self, capacity=None, num_steps=1):
...
+            state=state,
+            action=action,
+            reward=reward,
+            next_state=next_state,
+            next_action=next_action,
+            is_state_terminal=is_state_terminal,
+            **kwargs
+        )
+        last_n_transitions.append(experience)
         if is_state_terminal:
```
prabhatnagarajan: Why do we remove the if statement that checks if the transition is terminal?

tests/test_replay_buffer.py
```diff
@@ -487,6 +487,101 @@ def test_append_and_sample(self):
...
+        # 9 transitions for env_id=2
+        for i in range(9):
+            trans1 = dict(state=0, action=1, reward=2, next_state=3,
+                          next_action=4, is_state_terminal=False)
+            rbuf.append(env_id=2, **trans1)
+
+        # It should have:
+        #   - 4 transitions from env_id=1
+        #   - 5 transitions from env_id=2
+        self.assertEqual(len(rbuf), 9)
```
prabhatnagarajan: Shouldn't we have 3 transitions? One from `env_id=1` and five from `env_id=2`?

chainerrl/replay_buffer.py
muupan: i think it is not removed

tests/test_replay_buffer.py
muupan: Four form `env_id=1`: (s_0, s_4), (s_1, s_4), (s_2, s_4), and (s_3, s_4).

prabhatnagarajan: Got it.

PR #443, opened 2019-04-14
Add VectorStackFrame to reduce memory usage in train_dqn_batch_ale.py

muupan:
This PR adds

- `VectorEnvWrapper`: an VectorEnv analog to `gym.Wrapper`
- `VectorStackFrame`: an VectorEnv analog to `chainerrl.wrappers.atari_wrappers.StackFrame`

to reduce memory usage in `train_dqn_batch_ale.py`.

The current example wraps envs by `StackFrame` then by `MultiprocessVectorEnv`. This is actually memory inefficient because when `LazyFrames` is passed between processes via Pipe, it becomes not lazy anymore due to pickling. This PR solves the issue by stacking frames in the main process.

TODO:
- [x] add experimental results to validate this change

### DIFF
--- a/chainerrl/wrappers/vector_frame_stack.py
+++ b/chainerrl/wrappers/vector_frame_stack.py
@@ -78,11 +78,12 @@ class VectorFrameStack(VectorEnvWrapper):
         self.k = k
         self.stack_axis = stack_axis
         self.frames = [deque([], maxlen=k) for _ in range(env.num_envs)]
-        orig_shape = env.observation_space.shape
-        shape = list(orig_shape)
-        shape[self.stack_axis] *= k
+        orig_obs_space = env.observation_space
+        assert isinstance(orig_obs_space, spaces.Box)
+        low = np.repeat(orig_obs_space.low, k, axis=self.stack_axis)
+        high = np.repeat(orig_obs_space.high, k, axis=self.stack_axis)
         self.observation_space = spaces.Box(
-            low=0, high=255, shape=shape, dtype=np.uint8)
+            low=low, high=high, dtype=orig_obs_space.dtype)
 
     def reset(self, mask=None):
         batch_ob = self.env.reset(mask=mask)
--- a/chainerrl/wrappers/vector_frame_stack.py
+++ b/chainerrl/wrappers/vector_frame_stack.py
@@ -1,0 +1,107 @@
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from future import standard_library
+standard_library.install_aliases()  # NOQA
+
+from collections import deque
+
+from gym import spaces
+import numpy as np
+
+from chainerrl.wrappers.atari_wrappers import LazyFrames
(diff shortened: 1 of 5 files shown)


muupan:
I ran `python3 examples/ale/train_dqn_batch_ale.py --num-envs 8 --steps 10000000` for Breakout and SpaceInvaders with three random seeds each. `VectorFrameStack` and `FrameStack` corresponds to this PR and the master branch, respectively.

The achieved scores are similar.

image
image

Below are memory usage 10 minutes before the training finished (seed=0, Breakout). `VectorFrameStack` recuded memory usage a lot.
- `VectorFrameStack`: 9691M
- `FrameStack`: 31006M

chainerrl/wrappers/vector_frame_stack.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __str__(self):
+        return '<{}{}>'.format(type(self).__name__, self.env)
+
+    def __repr__(self):
+        return str(self)
+
+    @property
+    def unwrapped(self):
+        return self.env.unwrapped
```
keisuke-nakata: https://github.com/chainer/chainerrl/blob/master/chainerrl/envs/multiprocess_vector_env.py#L47
`MultiprocessVectorEnv` inherits `VectorEnv`, which does not have property `@unwrapped`.
It seems `vec_env` created at https://github.com/chainer/chainerrl/pull/443/files#diff-02e7d718fe75bb8ca034c551bd27ee05R166 will raise error when a user tries to access `vec_env.unwrapped`.

chainerrl/wrappers/vector_frame_stack.py
```diff
@@ -0,0 +1,107 @@
...
+        """Stack k last frames."""
+        VectorEnvWrapper.__init__(self, env)
+        self.k = k
+        self.stack_axis = stack_axis
+        self.frames = [deque([], maxlen=k) for _ in range(env.num_envs)]
+        orig_shape = env.observation_space.shape
+        shape = list(orig_shape)
+        shape[self.stack_axis] *= k
+        self.observation_space = spaces.Box(
+            low=0, high=255, shape=shape, dtype=np.uint8)
```
keisuke-nakata: For more generalization, `low=orig_obs_space.low, high=orig_obs_space.high, shape=shape, dtype=orig_obs_space.dtype` where `orig_obs_space = env.observation_space` looks better for me.

keisuke-nakata: I checked `chainerrl.wrappers.atari_wrappers.FrameStack` implementation.
It seems this makes `VectorFrameStack` equivalent to `chainerrl.wrappers.atari_wrappers.FrameStack`.
Please ignore comment above.

muupan: Actually I think your previous comment is right. It is better to follow the original observation space as much as possible. I'll fix it.

muupan:
Thanks for your review! I followed your suggestions.

PR #430, opened 2019-04-02
Precompute log probability in PPO

muupan:
~Merge #427 before this PR.~

Before this PR, for every minibatch (e.g. 64 transitions) in a batch (e.g. 2048 transitions) used for an iteration of PPO, both the current and old models are evaluated to compute the probability ratio. This PR instead pre-compute the outputs from the old model and evaluate only the current model for each minibatch. The reasons behind this PR are:
- It is computationally more efficient since it can use a larger batch and only evaluate the old model once per PPO iteration.
- Currently, `obs_normalizer` is updated after collecting a batch of transitions and before updating the policy. To correctly implement importance sampling, the policy used for collecting transitions should be used to compute importance weights. Thus, it is better to compute the probabilities before updating `obs_normalizer` which can affect the policy. https://github.com/chainer/chainerrl/blob/master/chainerrl/agents/ppo.py#L225
- It is what openai/baselines' ppo2 does.
  - compute and collect `neglogpacs` https://github.com/openai/baselines/blob/master/baselines/ppo2/runner.py#L33 
  - use them for updates for multiple epochs https://github.com/openai/baselines/blob/master/baselines/ppo2/ppo2.py#L135

This PR also adds `n_updates` to PPO's statistics to help debugging.

Todos
- [x] check experimental results on mujoco

PR #446, opened 2019-04-28
Recurrent and batched TRPO

muupan:
~Merge #431 before this PR.~

- Precompute action distributions for importance weighting and KL divergence as #430 did
- Support recurrent models as #431 did
- Support batch training


chainerrl/agents/trpo.py
```diff
@@ -350,27 +418,94 @@ def _update_policy(self, dataset):
...
+            policy_rs = self.policy.concatenate_recurrent_states(
+                [ep[0]['recurrent_state'][0] for ep in dataset])
+
+        flat_distribs, _ = self.policy.n_step_forward(
+            seqs_states, recurrent_state=policy_rs, output_mode='concat')
+
+        if self.standardize_advantages:
+            mean_advs = xp.mean(flat_advs)
+            std_advs = xp.std(flat_advs)
+            flat_advs = (flat_advs - mean_advs) / (std_advs + 1e-8)
```
toslunar: Non-recurrent code standardizes advantages before computing action distributions.  Could you make their order consistent?

muupan:
Let me investigate the failure in pfn-public-ci/chainerrl.py3.gpu .

muupan:
I added changes to address flexCI failurue:
- move `self.model` to gpu if policy is on gpu
- skip test cases with recurrent=True when either
  - gpu (because NStepLSTM with gpu does not support double backprop), or
  - chainer==7.0.0b3 (https://github.com/chainer/chainer/pull/8037)

Remaining flexCI failures are only flakey tests and #533

muupan:
@toslunar Can you review the changes I made?

PR #448, opened 2019-05-01
PPO example that reproduces the "Deep Reinforcement Learning that Matters" paper

muupan:
This PR adds a script to reproduce the PPO results in "Deep Reinforcement Learning that Matters" (http://arxiv.org/abs/1709.06560).

As you can see from the tables in README, our results are competitive with theirs. https://github.com/muupan/chainerrl/tree/ppo-mujoco-matters/examples/mujoco/ppo

To replicate the settings of the paper, I copied a new `Orthogonal` initializer from Chainer's PR https://github.com/chainer/chainer/pull/6031 to ChainerRL.


examples/mujoco/ppo/train_ppo.py
```diff
@@ -0,0 +1,199 @@
...
+    print('Action space:', action_space)
+
+    assert isinstance(action_space, gym.spaces.Box)
+
+    # Normalize observations based on their empirical mean and variance
+    obs_normalizer = chainerrl.links.EmpiricalNormalization(
+        obs_space.low.size, clip_threshold=5)
+
+    # Orthogonal weight initialization is used as OpenAI Baselines does
+    winit = chainerrl.initializers.Orthogonal(1.)
```
marioyc: These scale doesn't seem to be the same as the one used in the baseline's repo: https://github.com/openai/baselines/blob/master/baselines/common/models.py#L96
The other one seems okay though:
https://github.com/openai/baselines/blob/master/baselines/common/policies.py#L49

examples/mujoco/ppo/train_ppo.py
```diff
@@ -0,0 +1,199 @@
...
+from chainerrl.agents import PPO
+from chainerrl import experiments
+from chainerrl import misc
+
+
+def main():
+    import logging
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--gpu', type=int, default=0,
```
marioyc: Wouldn't the default be -1? or is 0 going to do the same?

examples/mujoco/ppo/train_ppo.py
```diff
@@ -0,0 +1,199 @@
...
+
+    agent = PPO(
+        model,
+        opt,
+        obs_normalizer=obs_normalizer,
+        gpu=args.gpu,
+        update_interval=args.update_interval,
+        minibatch_size=args.batch_size,
+        epochs=args.epochs,
+        clip_eps_vf=None,
```
marioyc: Baselines' PPO does seem to be clipping:
https://github.com/openai/baselines/blob/master/baselines/ppo2/defaults.py#L11
https://github.com/openai/baselines/blob/master/baselines/ppo2/model.py#L147
https://github.com/openai/baselines/blob/master/baselines/ppo2/model.py#L69

muupan:
Can you compare against their fork of baselines (linked from https://github.com/Breakend/DeepReinforcementLearningThatMatters), not the latest baselines?

examples/mujoco/ppo/train_ppo.py
muupan: `--gpu 0` uses GPU 0, while `--gpu -1` uses CPU only. Setting the default to CPU also makes sense, but currently all the example set the default to 0, GPU 0.

examples/mujoco/ppo/train_ppo.py
marioyc: Nevermind about the scale, this one seems to agree with:
https://github.com/Breakend/baselines/blob/50ffe01d254221db75cdb5c2ba0ab51a6da06b0a/baselines/ppo1/mlp_policy.py#L33

However the initializer appears to not be the same:
https://github.com/Breakend/baselines/blob/50ffe01d254221db75cdb5c2ba0ab51a6da06b0a/baselines/ppo1/mlp_policy.py#L28
https://github.com/Breakend/baselines/blob/50ffe01d254221db75cdb5c2ba0ab51a6da06b0a/baselines/common/tf_util.py#L294

examples/mujoco/ppo/train_ppo.py
marioyc: I see, no problem

examples/mujoco/ppo/train_ppo.py
marioyc: No problem here, previous version only used clipping on the policy network
https://github.com/Breakend/baselines/blob/50ffe01d254221db75cdb5c2ba0ab51a6da06b0a/baselines/ppo1/pposgd_simple.py#L115

marioyc:
Done, only difference seems to be on the initialization, theirs wouldn't require the orthogonal initializer

examples/mujoco/ppo/train_ppo.py
muupan: You are right, it seems they actually did not use orthogonal weight initialization at that time. I clarified it in the comment and README.

marioyc: Ok, everything else looks good to me

muupan:
CI failure seems due to #463 , so I merged this. Thanks for the review!

PR #449, opened 2019-05-01
TRPO example that reproduces the "Deep Reinforcement Learning that Matters" paper

muupan:
Merge ~#446~ ~#448~  first.

This PR adds a script to reproduce the TRPO results in "Deep Reinforcement Learning that Matters" (http://arxiv.org/abs/1709.06560).

As you can see from the tables in README, our results are competitive with theirs. https://github.com/muupan/chainerrl/tree/trpo-mujoco-matters/examples/mujoco/reproduction/trpo


examples/mujoco/reproduction/trpo/train_trpo.py
```diff
@@ -0,0 +1,189 @@
...
+    chainerrl.misc.draw_computational_graph(
+        [vf(fake_obs)], os.path.join(args.outdir, 'vf'))
+
+    # Hyperparameters in http://arxiv.org/abs/1709.06560
+    agent = chainerrl.agents.TRPO(
+        policy=policy,
+        vf=vf,
+        vf_optimizer=vf_opt,
+        obs_normalizer=obs_normalizer,
+        update_interval=args.trpo_update_interval,
```
toslunar: `max_kl=0.01,` might be inserted here, while it is the default.

examples/mujoco/reproduction/trpo/train_trpo.py
```diff
@@ -0,0 +1,189 @@
...
+    obs_space = env.observation_space
+    action_space = env.action_space
+    print('Observation space:', obs_space)
+    print('Action space:', action_space)
+
+    assert isinstance(obs_space, gym.spaces.Box)
+
+    # Normalize observations based on their empirical mean and variance
+    obs_normalizer = chainerrl.links.EmpiricalNormalization(
+        obs_space.low.size, clip_threshold=5)
```
toslunar: I didn't find `clip_threshold=5` in the paper, but `5` seems sufficiently large.

muupan: They used a fork of openai/baselines. TRPO uses `MlpPolicy`, which internally clip normalized observations by 5.
https://github.com/Breakend/baselines/blob/master/baselines/trpo_mpi/run_mujoco.py#L31
https://github.com/Breakend/baselines/blob/master/baselines/ppo1/mlp_policy.py#L25

PR #452, opened 2019-05-04
DDPG example that reproduces the TD3 paper

muupan:
This PR adds a script to reproduce the DDPG results in the TD3 paper (http://arxiv.org/abs/1802.09477).

As you can see from the tables in README, our results are competitive with theirs. https://github.com/muupan/chainerrl/tree/ddpg-mujoco-td3/examples/mujoco/ddpg

To replicate the settings of the TD3 paper's DDPG implementation (https://github.com/sfujim/TD3), I added following changes:
- `DDPG` agent now has `burnin_action_func` option that can be used to choose completely random actions before model is updated one or more times.
- `AdditiveGaussian` explorer now has `low` and `high` options. If specified, they are used to clip actions


examples/mujoco/ddpg/train_ddpg.py
```diff
@@ -0,0 +1,200 @@
...
+    timestep_limit = env.spec.tags.get(
+        'wrapper_config.TimeLimit.max_episode_steps')
+    obs_space = env.observation_space
+    action_space = env.action_space
+    print('Observation space:', obs_space)
+    print('Action space:', action_space)
+
+    action_size = action_space.low.size
+
+    winit = chainer.initializers.LeCunUniform(3 ** -0.5)
```
toslunar: Assuming the authors' implementation is https://github.com/sfujim/TD3/blob/master/OurDDPG.py, I guess the coefficient is 1.

toslunar: And it's equivalent with the current implementation in PyTorch
```python
init.kaiming_uniform_(self.weight, a=math.sqrt(5))
```
because the gain needed for the leaky ReLU with slope √5 is 1/√3.

toslunar:
imosCI, test this please.

PR #453, opened 2019-05-05
TD3 agent

muupan:
~Merge #452 before this PR.~

This PR adds a new agent named TD3 that implements the TD3 algorithm proposed by http://arxiv.org/abs/1802.09477.

As you can see from the tables in README, our results are competitive with theirs. https://github.com/muupan/chainerrl/tree/td3/examples/mujoco/td3

Not only TD3 but other recently added agents were missing in docs/agent.rst, so I synchronized it with `chainerrl/agents/__init__.py`.


chainerrl/agents/td3.py
```diff
@@ -0,0 +1,370 @@
...
+                self.explorer.select_action(
+                    self.t, lambda: batch_greedy_action[i])
+                for i in range(len(batch_greedy_action))]
+
+        self.batch_last_obs = list(batch_obs)
+        self.batch_last_action = list(batch_action)
+
+        return batch_action
+
+    def batch_observe_and_train(
```
prabhatnagarajan: Can you add documentation like here:  https://github.com/chainer/chainerrl/blob/master/chainerrl/agents/ddpg.py#L409

chainerrl/agents/td3.py
```diff
@@ -0,0 +1,370 @@
...
+                self.replay_buffer.append(
+                    state=self.batch_last_obs[i],
+                    action=self.batch_last_action[i],
+                    reward=batch_reward[i],
+                    next_state=batch_obs[i],
+                    next_action=None,
+                    is_state_terminal=batch_done[i],
+                )
+                if batch_reset[i] or batch_done[i]:
+                    self.batch_last_obs[i] = None
```
prabhatnagarajan: Shouldn't we have the `env_id` (https://github.com/chainer/chainerrl/blob/master/chainerrl/agents/ddpg.py#L442) since we might have an N-step replay buffer?

chainerrl/agents/td3.py
muupan: Since it is documented in https://github.com/chainer/chainerrl/blob/master/chainerrl/agent.py#L225, I don't think we need to repeat it here.

prabhatnagarajan: Ok.

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,201 @@
...
+    # Hyperparameters in http://arxiv.org/abs/1802.09477
+    agent = chainerrl.agents.TD3(
+        policy,
+        q_func1,
+        q_func2,
+        policy_optimizer,
+        q_func1_optimizer,
+        q_func2_optimizer,
+        rbuf,
+        gamma=0.99,
```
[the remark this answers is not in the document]
muupan: It would not help replicating the paper, because it would not change the behavior, right? Can you elaborate on how would it help?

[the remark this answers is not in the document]
prabhatnagarajan: It would not help in replicating the paper, but it would improve clarity, so that somebody using this as a reference knows it's a hyperparameter of importance.

chainerrl/agents/td3.py
muupan: Fixed.

examples/mujoco/td3/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+During each trial, the agent is trained for 1M timesteps and evaluated after every 5000 timesteps, resulting in 200 evaluations.
+Each evaluation reports average return over 10 episodes without exploration noise.
+
+### Max Average Return
+
+Maximum evaluation scores, averaged over 10 trials (+/- standard deviation), are reported for each environment.
+
+Reported scores are taken from the "TD3" column of Table 1 of [Addressing Function Approximation Error in Actor-Critic Methods](http://arxiv.org/abs/1802.09477).
```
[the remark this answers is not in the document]
muupan: Mentioned just below.

examples/mujoco/td3/README.md
```diff
@@ -0,0 +1,80 @@
...
+| Reacher-v2                |       **-2.55**+/-0.19 |          -3.60+/-0.56 |
+| InvertedPendulum-v2       |     **1000.00**+/-0.00 |    **1000.00**+/-0.00 |
+| InvertedDoublePendulum-v2 |      8435.33+/-2771.39 |   **9337.47**+/-14.96 |
+
+
+### Last 100 Average Return
+
+Average return of last 10 evaluation scores, averaged over 10 trials, are reported for each environment.
+
+Reported scores are taken from the "TD3" row of Table 2 of [Addressing Function Approximation Error in Actor-Critic Methods](http://arxiv.org/abs/1802.09477).
```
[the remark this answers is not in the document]
muupan: Mentioned just below.

examples/mujoco/td3/train_td3.py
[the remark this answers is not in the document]
muupan: I see, sure it may help to show that it is an important hyperprameter.

chainerrl/agents/td3.py
```diff
@@ -0,0 +1,370 @@
...
+    def update(self, experiences, errors_out=None):
+        """Update the model from experiences"""
+
+        batch = batch_experiences(experiences, self.xp, self.phi, self.gamma)
+        self.update_q_func(batch)
+        if self.q_func1_optimizer.t % self.policy_update_delay == 0:
+            self.update_policy(batch)
+            self.sync_target_network()
+
+    def select_greedy_action(self, obs):
```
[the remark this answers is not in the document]
muupan: I would not say calling it "greedy" is incorrect as it is trained to maximize Q as DDPG does, I renamed it to "onpolicy" as I think it is correct even when the policy is stochastic.

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,201 @@
...
+    args.outdir = experiments.prepare_output_dir(
+        args, args.outdir, argv=sys.argv)
+    print('Output files are saved in {}'.format(args.outdir))
+
+    # Set a random seed used in ChainerRL
+    misc.set_random_seed(args.seed, gpus=(args.gpu,))
+
+    def make_env(test):
+        env = gym.make(args.env)
+        # Unwrap TimiLimit wrapper
```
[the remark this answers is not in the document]
muupan: Fixed

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,202 @@
...
+                        help='GPU to use, set to -1 if no GPU.')
+    parser.add_argument('--load', type=str, default='',
+                        help='Directory to load agent from.')
+    parser.add_argument('--steps', type=int, default=10 ** 6,
+                        help='Total number of timesteps to train the agent.')
+    parser.add_argument('--eval-n-runs', type=int, default=10,
+                        help='Number of episodes run for each evaluation.')
+    parser.add_argument('--eval-interval', type=int, default=5000,
+                        help='Interval in timesteps between evaluations.')
+    parser.add_argument('--replay-start-size', type=int, default=10000,
```
prabhatnagarajan: Where did you get this number? I didn't see it in the paper. I know they do pure exploration for 10K frames, but I didn't see that they don't do any updates for 10K frames.

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,202 @@
...
+    timestep_limit = env.spec.tags.get(
+        'wrapper_config.TimeLimit.max_episode_steps')
+    obs_space = env.observation_space
+    action_space = env.action_space
+    print('Observation space:', obs_space)
+    print('Action space:', action_space)
+
+    action_size = action_space.low.size
+
+    winit = chainer.initializers.LeCunUniform(3 ** -0.5)
```
prabhatnagarajan: Where did you get this initializer? Not that these needs justification, but if it was a detail mentioned somewhere, then that's good to know. Otherwise, I'll just assume this initialization is safe.

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,202 @@
...
+        policy.xp.zeros_like(action_space.low, dtype=np.float32)[None],
+        name='action')
+    chainerrl.misc.draw_computational_graph(
+        [policy(fake_obs)], os.path.join(args.outdir, 'policy'))
+    chainerrl.misc.draw_computational_graph(
+        [q_func1(fake_obs, fake_action)], os.path.join(args.outdir, 'q_func1'))
+    chainerrl.misc.draw_computational_graph(
+        [q_func2(fake_obs, fake_action)], os.path.join(args.outdir, 'q_func2'))
+
+    rbuf = replay_buffer.ReplayBuffer(10 ** 6)
```
prabhatnagarajan: Nit: Technically the replay buffer stores the entire history, which in this specific case is 10 ** 6. Maybe you could set this to `args.steps`, though not a major deal.

examples/mujoco/td3/train_td3.py
```diff
@@ -0,0 +1,202 @@
...
+    action_size = action_space.low.size
+
+    winit = chainer.initializers.LeCunUniform(3 ** -0.5)
+
+    policy = chainer.Sequential(
+        L.Linear(None, 400, initialW=winit),
+        F.relu,
+        L.Linear(None, 300, initialW=winit),
+        F.relu,
+        L.Linear(None, action_size, initialW=winit),
```
prabhatnagarajan: On page 13/165, they say their policy outputs a a scalar? https://arxiv.org/pdf/1802.09477.pdf

prabhatnagarajan: The PR makes sense though...

chainerrl/agents/td3.py
```diff
@@ -0,0 +1,374 @@
...
+    def update(self, experiences, errors_out=None):
+        """Update the model from experiences"""
+
+        batch = batch_experiences(experiences, self.xp, self.phi, self.gamma)
+        self.update_q_func(batch)
+        if self.q_func1_optimizer.t % self.policy_update_delay == 0:
+            self.update_policy(batch)
+            self.sync_target_network()
+
+    def select_onpolicy_action(self, obs):
```
prabhatnagarajan: Is a separate `select_on_policy_action` necessary? Can this not be placed in `act` and `batch_act`?

chainerrl/agents/td3.py
```diff
@@ -0,0 +1,374 @@
...
+            action=self.last_action,
+            reward=reward,
+            next_state=state,
+            next_action=self.last_action,
+            is_state_terminal=done)
+
+        self.stop_episode()
+
+    def stop_episode(self):
+        self.last_state = None
```
prabhatnagarajan: Are we not supporting recurrence? E.g. like here: https://github.com/chainer/chainerrl/blob/master/chainerrl/agents/ddpg.py#L468?

chainerrl/agents/td3.py
muupan: It is just for readability and avoiding duplicate code.

chainerrl/agents/td3.py
muupan: This PR does not. Since the new interface of recurrent model is under review #431 , I think it is better to work on it later.

examples/mujoco/td3/train_td3.py
muupan: This is equivalent to PyTorch's default weight initialization. PyTorch is used by the original implementation: https://github.com/sfujim/TD3

examples/mujoco/td3/train_td3.py
muupan: This is what the original implementation does. https://github.com/sfujim/TD3/blob/master/TD3.py#L20

examples/mujoco/td3/train_td3.py
muupan: They used 10**6. https://github.com/sfujim/TD3/blob/master/utils.py#L8

examples/mujoco/td3/README.md
```diff
@@ -0,0 +1,82 @@
...
+Although the original paper used v1 versions of MuJoCo envs, we used v2 as v1 are not supported by recent versions of OpenAI Gym.
+
+| Environment               | ChainerRL Score | Reported Score |
+| ------------------------- |:---------------:|:--------------:|
+| HalfCheetah-v2            |     **9952.04** |        9532.99 |
+| Hopper-v2                 |     **3365.24** |        3304.75 |
+| Walker2d-v2               |     **4705.46** |        4565.24 |
+| Ant-v2                    |         4174.14 |    **4185.06** |
```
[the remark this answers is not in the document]
muupan: That's right. I'll add explanation on this.

examples/mujoco/td3/train_td3.py
muupan: They actually updated the model N times after every episode where N is the length of that episode. https://github.com/sfujim/TD3/blob/master/main.py#L94 So, this is different from what they did. I don't think this difference significantly affects, but I'll add explanation on this.

muupan:
Added the explanations. Thanks for the review!

PR #455, opened 2019-05-08
Fix a bug of unintentionally using same process indices

muupan:
I noticed some of batch training examples have a bug that unintentionally uses same process indices (thus same random seeds!) across env processes. This PR fixes the bug.

You can see the current and new behaviors by running the code below. 
```
import functools
import chainerrl
import gym

num_envs = 4


def make_env(process_idx, test):
    print(process_idx, test)
    return gym.make('Pendulum-v0')


def make_batch_env_old(test):
    return chainerrl.envs.MultiprocessVectorEnv(
        [(lambda: make_env(idx, test))
         for idx, env in enumerate(range(num_envs))])


def make_batch_env_new(test):
    return chainerrl.envs.MultiprocessVectorEnv(
        [functools.partial(make_env, idx, test)
         for idx, env in enumerate(range(num_envs))])


print('make_batch_env_old')
make_batch_env_old(test=False)
make_batch_env_old(test=True)

print('make_batch_env_new')
make_batch_env_new(test=False)
make_batch_env_new(test=True)
```

This code will outputs:
```
make_batch_env_old
3 False
3 False
3 False
3 False
3 True
3 True
3 True
3 True
make_batch_env_new
0 False
1 False
2 False
3 False
0 True
1 True
2 True
3 True
```

Even when same random seeds are used in env processes, actions sent by the agent are usually different due to  the stochasticity of policy or eplorer, so this may not result in noticeable difference in learning results, but this is definitely a bug that needs to be fixed.

### DIFF
--- a/examples/ale/train_a2c_ale.py
+++ b/examples/ale/train_a2c_ale.py
@@ -7,6 +7,7 @@ from future import standard_library
 standard_library.install_aliases()  # NOQA
 
 import argparse
+import functools
 import logging
 
 import chainer
@@ -123,7 +124,7 @@ def main():
 
     def make_batch_env(test):
         return chainerrl.envs.MultiprocessVectorEnv(
-            [(lambda: make_env(idx, test))
+            [functools.partial(make_env, idx, test)
              for idx, env in enumerate(range(args.num_envs))])
 
     sample_env = make_env(0, test=False)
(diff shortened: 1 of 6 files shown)


muupan:
Thanks for the review. I fixed `tests/wrappers_tests/test_vector_frame_stack.py` as well.

PR #460, opened 2019-05-22
fix ScaledFloatFrame.observation_space

keisuke-nakata:
`ScaledFloatFrame.observation_space` was not properly converted and it resulted in some inconsistency, such as `observation_space.low/high/sample/contains`.

### DIFF
--- a/chainerrl/wrappers/atari_wrappers.py
+++ b/chainerrl/wrappers/atari_wrappers.py
@@ -222,6 +222,7 @@ class ScaledFloatFrame(gym.ObservationWrapper):
     Especially, when the original env.observation_space is np.uint8,
     this wrapper converts frame values into [0.0, 1.0] of dtype np.float32.
     """
+
     def __init__(self, env):
         assert isinstance(env.observation_space, spaces.Box)
         gym.ObservationWrapper.__init__(self, env)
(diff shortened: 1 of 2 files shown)


muupan:
Can you fix flake8 and autopep8 errors?

keisuke-nakata:
Ah, I missed it. Will fix soon.

chainerrl/wrappers/atari_wrappers.py
```diff
@@ -217,13 +217,28 @@ def _get_ob(self):
...
     def __init__(self, env):
+        assert isinstance(env.observation_space, spaces.Box)
         gym.ObservationWrapper.__init__(self, env)
 
+        self.scale = 255.0
+
+        orig_obs_space = env.observation_space
+        self.observation_space = spaces.Box(
+            low=self._observation(orig_obs_space.low),
+            high=self._observation(orig_obs_space.high),
```
muupan: I think `low` and `high` should also be scaled, if not -inf or inf.

muupan: Actually it seems we don't need to consider whether it's inf or not since inf/255 is inf.

keisuke-nakata: Yes, `low` and `high` is scaled via `self._observation`.

muupan: Sorry I was confused. Nevermind.

PR #462, opened 2019-05-30
Support `gym>=0.12.2` by stopping to use underscore methods in gym wrappers

muupan:
- Stop using underscore methods in all the wrappers
- Require `pybullet>=2.4.9` for the grasping example because older ones do not work with `gym>=0.12.2`
- Revert the change requiring `gym<=0.12.1` by https://github.com/chainer/chainerrl/pull/461

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ chainer>=4.0.0
 fastcache; python_version<'3.2'
 funcsigs; python_version<'3.5'
 future
-gym>=0.9.7,<=0.12.1
+gym>=0.9.7
 numpy>=1.10.4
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ chainer>=4.0.0
 fastcache; python_version<'3.2'
 funcsigs; python_version<'3.5'
 future
-gym>=0.9.7,<=0.12.1
+gym>=0.9.7
```
toslunar: Note:
Looks good because the new method names are implemented in openai/gym#836 (0.9.6).

PR #465, opened 2019-06-03
Apply `noisy_net_sigma` parameter

keisuke-nakata:
The specified value for `noisy_net_sigma` was ignored when NoisyNet is enabled.


keisuke-nakata:
I fixed `examples/ale/train_dqn_batch_ale.py` as well.
(`examples/atari/rainbow/train_rainbow.py` was already fixed at the first commit)

PR #468, opened 2019-06-04
Code to collect demonstrations from an agent.

prabhatnagarajan:


chainerrl/experiments/collect_demos.py
```diff
@@ -0,0 +1,76 @@
...
+                info = {}
+            # a_t
+            a = agent.act(obs)
+            # o_{t+1}, r_{t+1}
+            new_obs, r, done, info = env.step(a)
+            # s_0, a_0, r__{t+1}, s_{t+1}
+            dataset.write((obs, a, r, new_obs, done, info))
+            obs = new_obs
+            reset = (done or episode_len == max_episode_len
+                     or info.get('needs_reset', False))
```
keisuke-nakata: `done` should be turned to `true` in this case if we follow [gym's implementation](https://github.com/openai/gym/blob/master/gym/wrappers/time_limit.py#L19) ?

chainerrl/experiments/collect_demos.py
```diff
@@ -0,0 +1,76 @@
...
+                done = False
+                r = 0
+                episode_r = 0
+                episode_len = 0
+                info = {}
+            # a_t
+            a = agent.act(obs)
+            # o_{t+1}, r_{t+1}
+            new_obs, r, done, info = env.step(a)
+            # s_0, a_0, r__{t+1}, s_{t+1}
```
keisuke-nakata: typo: `s_t, a_t, r_{t+1}, s_{t+1}`

chainerrl/experiments/collect_demos.py
prabhatnagarajan: I don't think so. I think `done` represents a terminal state. Sometimes you might end the episode even if it's not the terminal state. This information can affect the outcome of the algorithm.

keisuke-nakata: I understand. I'll close this conversation.

chainerrl/experiments/collect_demos.py
```diff
@@ -0,0 +1,76 @@
...
+            reset = (done or episode_len == max_episode_len
+                     or info.get('needs_reset', False))
+            timestep += 1
+            episode_len += 1
+            episode_r += r
+            episode_num = episode_num + 1 if reset else episode_num
+            if steps is None:
+                terminate = episode_num >= episodes
+                # log final episode
+                logger.info('demonstration episode %s length:%s R:%s',
```
keisuke-nakata: this logger will be triggered at each step?

chainerrl/experiments/collect_demos.py
prabhatnagarajan: Will fix.

chainerrl/experiments/collect_demos.py
prabhatnagarajan: Yes, this is an error. I should be logging the final episode. I'll modify.

examples/ale/collect_demos_ale.py
```diff
@@ -0,0 +1,126 @@
...
+    parser.add_argument('--env', type=str, default='BreakoutNoFrameskip-v4',
+                        help='OpenAI Atari domain to perform algorithm on.')
+    parser.add_argument('--outdir', type=str, default='results',
+                        help='Directory path to save output files.'
+                             ' If it does not exist, it will be created.')
+    parser.add_argument('--seed', type=int, default=0,
+                        help='Random seed [0, 2 ** 31)')
+    parser.add_argument('--gpu', type=int, default=0,
+                        help='GPU to use, set to -1 if no GPU.')
+    parser.add_argument('--demo', action='store_true', default=False)
```
keisuke-nakata: `demo` is not used. (always `true` in this script)

examples/ale/collect_demos_ale.py
```diff
@@ -0,0 +1,126 @@
...
+        [q_func(np.zeros((4, 84, 84), dtype=np.float32)[None])],
+        os.path.join(args.outdir, 'model'))
+
+    # Use the same hyperparameters as the Nature paper
+    opt = optimizers.RMSpropGraves(
+        lr=2.5e-4, alpha=0.95, momentum=0.0, eps=1e-2)
+
+    opt.setup(q_func)
+
+    rbuf = replay_buffer.ReplayBuffer(10 ** 6)
```
keisuke-nakata: It depends on a situation, but I think we should clarify that `opt/rbuf` are dummies. (but `agent` requires them)

keisuke-nakata:
CI fails 😢

prabhatnagarajan:
Merge https://github.com/chainer/chainerrl/pull/481 before this one.

examples/atari/collect_demos_ale.py
```diff
@@ -0,0 +1,119 @@
...
+    print('Output files are saved in {}'.format(args.outdir))
+
+    def make_env():
+        env = atari_wrappers.wrap_deepmind(
+            atari_wrappers.make_atari(args.env, max_frames=None),
+            episode_life=False,
+            clip_rewards=False)
+        env.seed(int(args.seed))
+        # Randomize actions like epsilon-greedy
+        env = chainerrl.wrappers.RandomizeAction(env, 0.01)
```
keisuke-nakata: Is this randomization intentional?

examples_tests/atari/test_collect_demos.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -Ceu
+
+outdir=$(mktemp -d)
+
+gpu="$1"
+
+# Chainer 4 does not support open_pickle_dataset_writer, which is used by the demonstration collection example.
+pickle_writer_support=$(python -c "import chainer; from distutils.version import StrictVersion; print(1 if StrictVersion(chainer.__version__) >= StrictVersion('5.0.0') else 0)")
```
keisuke-nakata: On some environment this line may not work because users might use `python3` instead of `python`.  
I think now this line can be removed because we no longer support Chainer V4.

examples/atari/collect_demos_ale.py
prabhatnagarajan: Yes. We wouldn't want to collect demonstrations that are identical in a deterministic env.

examples_tests/atari/test_collect_demos.sh
prabhatnagarajan: I don't think the Chainer v4 support has been dropped yet. That will be done separately in a different PR.

Regarding your other comment, I'm following this: https://github.com/chainer/chainerrl/pull/204/files#diff-875175ed9e296e37b03562cf27435d32R11.

And wouldn't your `python2` vs. `python3` comment apply to all example tests? E.g. https://github.com/chainer/chainerrl/blob/master/examples_tests/atari/test_a2c.sh

keisuke-nakata: Ah you're completely right.

PR #473, opened 2019-06-09
Drop python2 support

### LINKED ISSUE #392: Support Python2/3 JSON dumps of dicts.

### LINKED ISSUE #463: CI failure with Python 2 and chainer==6.0.0
This will be solved by https://github.com/chainer/chainer/pull/7323

### LINKED ISSUE #467: Drop python2 support
Even numpy stopped too support python 2 from January 1, 2019. (https://www.numpy.org/neps/nep-0014-dropping-python2.7-proposal.html)

muupan:
~Resolves #463~
Resolves #392
Resolves #467 

- Travis CI with python 2 is stopped
- `future` is removed from dependencies
- import from `__future__` is removed
- hacks for python 2 are removed (#392)
- switched from `ppa:jonathanf` to `ppa:cran` for installing ffmpeg since the former is no longer available
- remove use of six
- remove fastcache, funcsigs, statistics as dependencies


chainerrl/optimizers/nonbias_weight_decay.py
```diff
@@ -1,11 +1,4 @@
 # This caused an error in py2 because cupy expect non-unicode str
```
marioyc: Seems like this line should also be removed

muupan:
Good point! I removed six. By this change`chainerrl.misc.makedirs` is now completely the same as os.makedirs, so I removed the file.

requirements.txt
```diff
@@ -2,7 +2,6 @@ cached-property
 chainer>=4.0.0
 fastcache; python_version<'3.2'
```
marioyc: Since this is only necessary for Python below 3.2 and we say ChainerRL is tested with 3.5+, shouldn't we remove this?
I guess something similar applies to statistics.

marioyc: This import would need to be changed accordingly for fastcache: https://github.com/chainer/chainerrl/blob/e424a75c3c7560fb6d78594e70f52af9668788fb/chainerrl/functions/lower_triangular_matrix.py#L12

And here for funcsigs: https://github.com/chainer/chainerrl/blob/7523bb7a7f6ec4e4236875b6990837a9e7dcfd13/chainerrl/links/sequence.py#L13

PR #476, opened 2019-06-11
Add warning about numpy 1.16.0

### LINKED ISSUE #475: Warn if numpy is 1.16.0
`numpy==1.16.0` causes memory leak with `MultiprocessVectorEnv` because ndarrays are pickled when going through pipes. See https://github.com/numpy/numpy/issues/12793

muupan:
Resolves #475

### DIFF
--- a/chainerrl/envs/multiprocess_vector_env.py
+++ b/chainerrl/envs/multiprocess_vector_env.py
@@ -9,6 +9,7 @@ standard_library.install_aliases()  # NOQA
 from multiprocessing import Pipe
 from multiprocessing import Process
 import signal
+import warnings
 
 from cached_property import cached_property
 import numpy as np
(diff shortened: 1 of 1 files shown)


muupan:
yes

chainerrl/envs/multiprocess_vector_env.py
```diff
@@ -53,6 +54,13 @@ class MultiprocessVectorEnv(chainerrl.env.VectorEnv):
     """
 
     def __init__(self, env_fns):
+        if np.__version__ == '1.16.0':
```
toslunar: In general, `numpy.lib.NumpyVersion` could be used.  Looks OK without it to detect a single version.

muupan: I intend to detect a single version. I haven't check if pre-release versions of numpy have the same issue.

PR #478, opened 2019-06-12
Add CI configs

imos:
Currently, some imosCI test targets are running, but I'd like to stop them and add the following targets instead:
- https://ci.preferred.jp/chainerrl.py3.gpu/
- https://ci.preferred.jp/chainerrl.py3.cpu/
- https://ci.preferred.jp/chainerrl.py3.chainer4/

The details for each test target should be written in .pfnci/config.pbtxt.

### DIFF
--- a/.pfnci/hint.pbtxt
+++ b/.pfnci/hint.pbtxt
@@ -1,0 +1,16 @@
+# hint.pbtxt is a config file for xpytest.
+#
+# Proto type: xpytest.proto.HintFile
+# https://github.com/chainer/xpytest/blob/master/proto/test_case.proto
+
+# Slow tests take 60+ seconds.
+rules { name: "agents_tests/test_ppo.py" xdist: 4 deadline: 240 }
+rules { name: "agents_tests/test_trpo.py" xdist: 4 deadline: 240 }
+
+# Slow tests take 10+ seconds.
+rules { name: "agents_tests/test_acer.py" }
+rules { name: "agents_tests/test_ale.py" }
+rules { name: "agents_tests/test_trpo.py" }
+rules { name: "policies_tests/test_deterministic_policy.py" }
+rules { name: "q_functions_tests/test_state_action_q_function.py" }
+rules { name: "tests/test_ale.py" }
--- a/.pfnci/run.sh
+++ b/.pfnci/run.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+# run.sh is a script to run unit tests inside Docker.  This should be injected
+# by and called from script.sh.
+#
+# Usage: run.sh [target]
+# - target is a test target (e.g., "py37").
+#
+# Environment variables:
+# - GPU (default: 0) ... Set a number of GPUs to GPU.
+#       CAVEAT: Setting GPU>0 disables non-GPU tests, and setting GPU=0 disables
+#               GPU tests.
+
+set -eux
+
+cp -a /src /chainerrl
+mkdir -p /chainerrl/.git
+cd /
+
+# Remove pyc files.  When the CI is triggered with a user's local files, pyc
+# files generated on the user's local machine and they often cause failures.
+find /chainerrl -name "*.pyc" -exec rm -f {} \;
+
+TARGET="$1"
+: "${GPU:=0}"
+: "${XPYTEST_NUM_THREADS:=$(nproc)}"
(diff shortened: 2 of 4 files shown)


imos:
FYI, the commit's test results are available at:
https://ci.preferred.jp/p/dashboard_by_commit_id?commit_id=d9b6c21c09fa132073f8f8d664649009d4ae6831

.pfnci/hint.pbtxt
```diff
@@ -0,0 +1,16 @@
...
+rules { name: "agents_tests/test_ppo.py" xdist: 4 deadline: 240 }
+rules { name: "agents_tests/test_trpo.py" xdist: 4 deadline: 240 }
+
+# Slow tests take 10+ seconds.
+rules { name: "agents_tests/test_acer.py" }
+rules { name: "agents_tests/test_ale.py" }
+rules { name: "agents_tests/test_trpo.py" }
+rules { name: "policies_tests/test_deterministic_policy.py" }
+rules { name: "q_functions_tests/test_state_action_q_function.py" }
+rules { name: "tests/test_ale.py" }
```
muupan: これってnameだけ指定してるときにどういう意味があるかわからないんですが、xpytestのどこを見ればわかりますか？

imos: Created a pull request in xpytest to clarify the situation: https://github.com/chainer/xpytest/pull/10

muupan: Sorry but I still don't understand what `rules { name: "agents_tests/test_acer.py" }` does.

imos: This changes the order of test cases to test (the specified tests are prioritized). It enables xpytest to use CPU cores more efficiently because long lingering tests often make some CPU cores idle.

muupan: My understanding is that tests specified in `.pfnci/hint.pbtxt` will be executed earlier than the non-specified, so tests that take long time should be specified here to utilize CPUs efficiently. That makes sense. Correct if I'm wrong.

.pfnci/run.sh
```diff
@@ -0,0 +1,85 @@
...
+  if [ "${CHAINER}" != '' ]; then
+    "${PYTHON}" -m pip install "chainer==${CHAINER}"
+  fi
+
+  if [ "${PYTHON}" == 'python' ]; then
+    "${PYTHON}" -m pip install \
+        'cython==0.28.0' 'numpy<1.10' 'scipy<0.19' 'more-itertools<=5.0.0'
+  fi
+
+  # TODO(chainerrl): Enable ChainerRL to work without -e.
```
muupan: Does current ChainerRL have any problem without `-e`?

imos: I don't know what happens exactly, but https://ci.preferred.jp/chainerrl.py3.cpu/17285/ is the result of running without `-e`. Probably, some files are not exported correctly.

muupan: The error message `TypeError: train_agent_async() got an unexpected keyword argument 'eval_n_steps'` seems to imply that it uses an old version of chainerrl.

muupan: I successfully run the test without `-e`. The only change from d9b6c21c09fa132073f8f8d664649009d4ae6831 is removing `-e`, so I'm not sure why it failed when you tried. https://github.com/chainer/chainerrl/pull/478

muupan:
```
diff --git a/.pfnci/run.sh b/.pfnci/run.sh
index c3d16eb..c7a7b77 100644
--- a/.pfnci/run.sh
+++ b/.pfnci/run.sh
@@ -68,8 +68,7 @@ main() {
         'cython==0.28.0' 'numpy<1.10' 'scipy<0.19' 'more-itertools<=5.0.0'
   fi

-  # TODO(chainerrl): Enable ChainerRL to work without -e.
-  "${PYTHON}" -m pip install -e /chainerrl
+  "${PYTHON}" -m pip install  /chainerrl
   # TODO(chainerrl): Prepare test target instead.
   "${PYTHON}" -m pip install \
       'pytest==4.1.1' 'pytest-xdist==1.26.1' 'mock' \
```
This is the diff. Can you check it again?

imos: I'm sorry, but I found my repository has cache of egg files... It is resolved.

PR #480, opened 2019-06-13
[WIP] DQfD

ummavi:
Follows https://github.com/chainer/chainerrl/pull/468.
WIP implementation of DQfD

TODO: 
* Implement a replay buffer that does proportional sampling across both demonstration buffer and self-play buffer
* Benchmark on ATARI
* Test with batch-mode DQN
* Test cases


chainerrl/agents/dqfd.py
```diff
@@ -0,0 +1,620 @@
...
+                                      batch_states=self.batch_states)
+
+        exp_batch['weights'] = self.xp.asarray(
+            [elem[0]['weight']for elem in experiences], dtype=self.xp.float32)
+
+        errors_out = []
+        loss_q_nstep, loss_q_1step = self._compute_ddqn_losses(
+            exp_batch, errors_out=errors_out)
+
+        # Add the agent/demonstration bonus priorities and update
```
prabhatnagarajan: Should we be updating the priorities even during pretraining? Because it looks like that's what's happening.

examples/atari/dqfd/train_dqfd.py
```diff
@@ -0,0 +1,264 @@
...
+        args.replay_buffer_size, alpha=0.4,
+        beta0=0.6, betasteps=betasteps,
+        num_steps=args.num_step_return)
+
+    # Fill the demo buffer with expert transitions
+    n_demo_transitions = 0
+    with chainer.datasets.open_pickle_dataset(args.expert_demo_path) as dset:
+        for transition in dset:
+            (obs, a, r, new_obs, done, info) = transition
+            # Assuming r is already normalized to log-scale.
```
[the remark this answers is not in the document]
ummavi: Yes, it is. Would it be clearer to undo it during the dataset creation and move the logic here?

chainerrl/agents/dqfd.py
ummavi: This is unclear from the paper. After some discussion, we decided it makes more sense to not have it update during pre-training. I'll make this change.

examples/atari/dqfd/train_ddqn.py
```diff
@@ -0,0 +1,221 @@
...
+    n_actions = env.action_space.n
+    q_func = DuelingDQN(n_actions)
+
+    # Draw the computational graph and save it in the output directory.
+    chainerrl.misc.draw_computational_graph(
+        [q_func(np.zeros((4, 84, 84), dtype=np.float32)[None])],
+        os.path.join(args.outdir, 'model'))
+
+    # Use the same hyperparameters as the Nature paper
+    opt = optimizers.RMSpropGraves(
```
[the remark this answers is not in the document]
ummavi: They do not. I assumed this since they take a lot of the parameters from the double-DQN paper, which states:
```The optimization employed to train the network is RMSProp (with momentum parameter 0.95).```
Does this indeed correspond to optimizers.RMSpropGraves with alpha=0.95?

File: chainerrl/agents/dqfd.py [hunk 4]
```diff
@@ -0,0 +1,679 @@
...
+                                         normalize_by_max=normalize_by_max,
+                                         error_min=error_min,
+                                         error_max=error_max,
+                                         num_steps=num_steps)
+
+        self.memory = PrioritizedBuffer(capacity)
+        self.memory_demo = PrioritizedBuffer(None)
+
+    def weights_from_probabilities(self, probabilities, min_probability):
+        """Overwrite weights_from_probabilities to make beta increment explicit
```
prabhatnagarajan: I think it would be better to not overwrite `weights_from_probabilities`, and instead modify the `PriorityWeightError`'s `weights_from_probabilities` to also take in a bool `update_beta`, which defaults to `True`. Then, in your `sample` function, you can take in a bool `pretraining`, which is then passed to `weights_to_probabilities` as `update_beta=pretraining`. That way, you can remove  the `update_beta` function, and in your `ReplayUpdater`, you can take in a flag `pretraining` that's passed through `update_if_necessary`, which can then be used when sampling transitions. Give this some thought. I think this drastically reduces duplicate code, sticks more with convention, and reduces likelihood of bugs.

chainerrl/agents/dqfd.py
```diff
@@ -0,0 +1,679 @@
...
+            clip_delta=self.clip_delta,
+            batch_accumulator=self.batch_accumulator)
+        loss_nstep = compute_weighted_value_loss(
+            y, t, exp_batch['weights'],
+            mask=exp_batch["is_n_step"],
+            clip_delta=self.clip_delta,
+            batch_accumulator=self.batch_accumulator)
+        return loss_nstep, loss_1step
+
+    def act_and_train(self, obs, reward):
```
marioyc: Seems like inheriting this from DQN would do since there doesn't seem to be any change

chainerrl/agents/dqfd.py
```diff
@@ -0,0 +1,679 @@
...
+            ('average_loss_1step', self.average_loss_1step),
+            ('average_loss_nstep', self.average_loss_nstep),
+            ('average_loss_supervised', self.average_loss_supervised),
+            ('average_loss', self.average_loss),
+            ('n_updates', self.optimizer.t),
+            ('average_agent_td_error', self.average_agent_td_error),
+            ('average_demo_td_error', self.average_demo_td_error)
+        ]
+
+    def batch_observe_and_train(self, batch_obs, batch_reward,
```
marioyc: Same as act_and_train

chainerrl/agents/dqfd.py
ummavi: Absolutely right. Thanks!

chainerrl/agents/dqfd.py
ummavi: I tried doing this but since `weights_from_probabilities` is called twice- one for each buffer, I'll have to call one with `update_beta=True` and another with `update_beta=False` which is not so nice. I'll see if I can come up with a better way.

PR #483, opened 2019-06-14
Moves replay buffers to a directory

prabhatnagarajan:
As the number of different types of replay buffers grows, it makes more sense to have a dedicated directory for the different types of replay buffers. This PR does that, while maintaining backwards compatibility.


prabhatnagarajan:
I ported all the the replay buffers into a separate directory, maintaining backwards compatibility. I'd like to change the internal imports in a separate PR. This should be straightforward to review. I simply copied over replay buffers to separate files. I also include the following recent PR: https://github.com/chainer/chainerrl/pull/485/files (https://github.com/chainer/chainerrl/pull/483/commits/cc3e7e0b09027482b749ae524bf096c9efe731dd), so the copied files are up to date.

PR #491, opened 2019-06-27
Monitor with ContinuingTimeLimit support

keisuke-nakata:
The original gym's `Monitor` assumes the "end of episode" as `done=True`, but ChainerRL's `ContinuingTimeLimit` forces agents reset the env even if `done=False`.


chainerrl/wrappers/continuing_time_limit.py
```diff
@@ -46,3 +51,76 @@ def step(self, action):
...
         self._elapsed_steps = 0
         return self.env.reset()
+
+
+class ContinuingTimeLimitMonitor(Monitor):
+    """`Monitor` with ChainerRL's `ContinuingTimeLimit` support.
+
+    Because of the original implementation's design,
+    explicit `close()` is needed to save the last episode.
+    Do not forget to call `close()` at the last line of your script.
```
muupan: As I understand correctly, `gym.wrappers.Monitor`'s `close` is always called automatically by
https://github.com/openai/gym/blob/master/gym/wrappers/monitor.py#L92. Doesn't it work for our case?

chainerrl/wrappers/continuing_time_limit.py
```diff
@@ -46,3 +51,76 @@ def step(self, action):
...
+    For details, see
+    https://github.com/openai/gym/blob/master/gym/wrappers/monitor.py
+    """
+
+    def _start(self, directory, video_callable=None, force=False, resume=False,
+               write_upon_reset=False, uid=None, mode=None):
+        if self.env_semantics_autoreset:
+            raise gym.error.Error(
+                "Detect 'semantics.autoreset=True' in `env.metadata`, "
+                "which means the env comes from deprecated OpenAI Universe.")
```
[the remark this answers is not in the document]
keisuke-nakata: > OpenAI Universe is deprecated, but Gym still supports `semantics.autoreset=True`, correct?

I think so.

> If so, I think we should state here that `ContinuingTimeLimitMonitor` does not support it.

OK, I'll add doc for that.

> In addition, I think we should not use `gym.error.Error` since it is us, not gym, who raises that error.

I think we can use `gym.error.Error` because `ContinuingTimeLimitMonitor` inherits  `gym.Monitor`, but it is okay to raise `ValueError` or `NotImplementedError` for this `if` clause in order to clarify that we do not support Universe.

chainerrl/wrappers/continuing_time_limit.py
keisuke-nakata: I didn't know that. I'll update the doc

PR #497, opened 2019-07-03
Adds training times for reproduced Mujoco results

prabhatnagarajan:
Algorithm times added in this PR:

- [x] TD3
- [x] PPO
- [x] DDPG


ummavi:
I think the first table for TD3 looks a bit off in rich view mode.

muupan:
Information about machines would be needed for training times to be useful.

prabhatnagarajan:
https://github.com/prabhatnagarajan/chainerrl/tree/mujoco_times/examples/mujoco/reproduction

PR #509, opened 2019-07-17
Improve the algorithm list on README

muupan:
This PR improve the algorithm list on README by adding links to papers and examples. The paper list is removed. You can check the new README here: https://github.com/chainer/chainerrl/tree/improve-algorithm-list

I removed `PGT` because it does not follow any paper and I think it's not worth mentioning in README. I'm not sure if the technique list covers all the techniques worth mentioning, but we can add later.


README.md
```diff
@@ -52,41 +52,44 @@ For more information, you can refer to [ChainerRL's documentation](http://chaine
...
 Following algorithms have been implemented in ChainerRL:
-- A3C (Asynchronous Advantage Actor-Critic)
-- ACER (Actor-Critic with Experience Replay)
-- Asynchronous N-step Q-learning
-- Rainbow
-- Categorical DQN
-- IQN
-- DQN (including Double DQN, Persistent Advantage Learning (PAL), Double PAL, Dynamic Policy Programming (DPP))
-- DDPG (Deep Deterministic Policy Gradients) (including SVG(0))
-- PGT (Policy Gradient Theorem)
```
prabhatnagarajan: We're not going to include PGT in the new README?

muupan: I wrote why I removed it in the PR.

PR #510, opened 2019-07-18
Use chainer.as_variable

muupan:
We no longer need define our own function to wrap an array by a variable as `chainer` has `as_variable` since v4 (https://docs.chainer.org/en/v4.0.0/reference/generated/chainer.as_variable.html).

### DIFF
--- a/chainerrl/distribution.py
+++ b/chainerrl/distribution.py
@@ -19,13 +19,6 @@ import numpy as np
 from chainerrl.functions import mellowmax
 
 
-def _wrap_by_variable(x):
-    if isinstance(x, chainer.Variable):
-        return x
-    else:
-        return chainer.Variable(x)
-
-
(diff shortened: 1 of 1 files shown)


toslunar:
Do the args require grads?

muupan:
Good point. It is difficult to guarantee that we won't compute gradient wrt `mean` or `var` of `GaussianDistribution`, so maybe we should not use `chainer.as_variable` here...

muupan:
Some facts
- `ACER` computes gradient wrt Distribution.params via `backward`, not `chainer.grad`, so it can be affected by `requires_grad=False`.
- Currently `chainer.grad` can compute gradient even when `requires_grad=False`, but this is a bug of Chainer and will be fixed (https://github.com/chainer/chainer/issues/5974)

PR #511, opened 2019-07-18
Use chainer.grad in ACER

muupan:
It is simpler than `backward` with `backprop_truncated`.

Todo
- [x] check training performance on atari


muupan:
I confirmed this change does not affect training performance.

`python3 examples/atari/train_acer_ale.py 16` was run for three random seeds for each game.

BreakoutNoFrameskip-v4
SeaquestNoFrameskip-v4
SpaceInvadersNoFrameskip-v4

PR #520, opened 2019-07-31
Specify ubuntu 16.04 for Travis CI and modify a dependency accordingly

### LINKED ISSUE #519: Travis CI failure with ubuntu 16.04
Since we don't specify which distribution to use, Travis CI seems to sometimes use Ubuntu 16.04 and sometimes 14.04. However, the apt repository added by #491 seems to work only with 14.04, which causes CI failure https://travis-ci.org/chainer/chainerrl/jobs/565824376.

Now that 14.04 is no longer supported, I think we should explicitly specify 16.04 and make it work with it.

muupan:
Resolves #519.

- Specify ubuntu 16.04 (xenial) as a distribution to use in Travis CI. See #519 for the reason.
- Use xvfb service since the previous way does not work with 16.04.
- Install `fleeglut3-dev` to avoid `ImportError: Library "GLU" not found.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ services:
 before_install:
   - sudo add-apt-repository -y ppa:jonathonf/ffmpeg-4
   - sudo apt-get update
+  - sudo apt install -y freeglut3-dev ffmpeg
 # command to install dependencies
 install:
-  - sudo apt install -y freeglut3-dev ffmpeg
   - pip install --upgrade pip setuptools wheel
   - |
     if [[ $CHAINER_VERSION == 4 ]]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,18 @@
...
   - "2.7"
   - "3.6"
 env:
   - CHAINER_VERSION=4
   - CHAINER_VERSION=stable
+services:
+  - xvfb
 before_install:
-  - sudo add-apt-repository -y ppa:mc3man/trusty-media
+  - sudo add-apt-repository -y ppa:jonathonf/ffmpeg-4
```
prabhatnagarajan: Why did you change these lines?

PR #538, opened 2019-08-26
Avoid cupy.zeros_like with numpy.ndrray

muupan:
becuase cupy started to raise an error from cupy==6.0.0.


examples/mujoco/train_trpo_gym.py
```diff
@@ -156,7 +156,7 @@ def make_env(test):
 
     # Draw the computational graph and save it in the output directory.
     fake_obs = chainer.Variable(
-        policy.xp.zeros_like(obs_space.low, dtype=np.float32)[None],
+        policy.xp.zeros(obs_space.low.shape, dtype=np.float32)[None],
```
toslunar: LGTM because the `gym.spaces.Box.low` attribute seems a `numpy.ndarray`.

PR #541, opened 2019-08-28
fix function call

marioyc:


tests/test_distribution.py
```diff
@@ -100,6 +100,12 @@ def test_log_prob(self):
             np.testing.assert_almost_equal(np.log(batch_p.array),
                                            batch_log_p.array)
 
+    def test_sample_with_log_prob(self):
+        for i in range(self.n):
```
muupan: `self.n` is the number of categories modeled by the distribution, but here it is used just as the number of repeats, and `i` is not used inside the loop. Can you use a constant number as in `TestGaussianDistribution` instead?

tests/test_distribution.py
```diff
@@ -164,6 +170,12 @@ def test_log_prob(self):
...
```
muupan: Ditto.

tests/test_distribution.py
marioyc: Yes, I used self.n becaused then in expectation you'll see each number once when sampling, but I can change it to a constant

muupan: Ah, makes sense! Sorry, then it is ok to keep it as it is. Can you add a comment to explain it instead?

marioyc: Done

PR #560, opened 2019-09-19
Fixes Rainbow Score to use correct Reporting Protocol

prabhatnagarajan:
Currently, the Rainbow scores show the scores of the final agent, as opposed to a re-evaluation of the best intermediate network.

The new README looks like this: https://github.com/prabhatnagarajan/chainerrl/tree/fixed_scores/examples/atari/reproduction/rainbow


muupan:
Did you include `examples/pretrained/` by mistake?

Issue #551, opened 2019-09-08
Add trained models


prabhatnagarajan:
- Add an example where a user can load a trained model.

Pit-Storm:
There are trained Models now, but one is not able to download it. If I run the reproduction example for A3C with pretrained model demo, I got a HTTP Error "403: Forbidden" for the URL  https://chainer-assets.preferred.jp/chainerrl/A3C/Breakout\final.zip

I think this is a convenience with joining paths. Are you joining the URL-Path with os.path.join?
When it's coming to on-disk-paths python can handle \ or / in whatever OS it is. On windows os.path.join joins with \ and on unix systems it joins with /. When you join a URL on windows it joins a \ instead a /.

When i replace the \ (right before final.zip in the upper URL) with an / I can access the file manually.

Maybe you can fix it in your code?

prabhatnagarajan:
You're correct that we use join: https://github.com/chainer/chainerrl/blob/master/chainerrl/misc/pretrained_models.py#L124
`cache_path = cached_download(os.path.join(url_basepath,file))`
Where `file` is `final.zip`. 

Thanks for sharing! I'll try and amend this.

PR #565, opened 2019-10-02
Releases trained models for all reproduced agents

prabhatnagarajan:
Addresses: [#551](https://github.com/chainer/chainerrl/issues/551)


muupan:
Can you fix the Travis CI errors?

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,149 @@
...
+    It downloads a file from the URL if there is no corresponding cache. After
+    the download, this function stores a cache to the directory under the
+    dataset root (see :func:`set_dataset_root`). If there is already a cache
+    for the given URL, it just returns the path to the cache without
+    downloading the same file.
+    Args:
+        url (string): URL to download from.
+    Returns:
+        string: Path to the downloaded file.
+    """
```
muupan: - Why did you remove the empty lines in the docstring? We use Google Style docstrings http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings
- It seems like you added some change to this function, e.g., 'CHAINERRL_DOWNLOAD_REPORT'. Can you change the docstring accordingly and highlight the changes from ChainerCV's function?

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,149 @@
...
+def download_and_store_model(alg, url, env, model_type):
+    """Downloads a model file and puts it under model directory.
+    It downloads a file from the URL and puts it under model directory.
+    If there is already a file at the destination path,
+    it just returns the path without downloading the same file.
+    Args:
+        url (string): URL to download from.
+    Returns:
+        string: Path to the downloaded file.
+    """
```
muupan: Ditto

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,149 @@
...
+    "Rainbow": ["best", "final"],
+    "A3C": ["final"],
+    "DDPG": ["best", "final"],
+    "TRPO": ["best", "final"],
+    "PPO": ["final"],
+    "TD3": ["best", "final"],
+    "SAC": ["best", "final"]
+}
+
+url = "https://chainer-assets.preferred.jp/chainerrl/"
```
muupan: This module-level variable should be renamed as `cached_download` and `download_and_store_model` use a local variable of the same name.

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,149 @@
...
+        if not is_cached:
+            cache_path = cached_download(os.path.join(url_basepath,
+                                                      file))
+            os.rename(cache_path, path)
+            with zipfile.ZipFile(path, 'r') as zip_ref:
+                zip_ref.extractall(root)
+        return os.path.join(root, model_type), is_cached
+
+
+def download_model(alg, env, model_type="best"):
```
muupan: Can you add a docstring?

prabhatnagarajan:
To test this locally:

For each of the 9 run reproducibility scripts, run:

`python train_<alg>.py --load-pretrained --pretrained-type <final|best> --env <Env>`

Note that for some algorithms we don't have both `final` and `best` networks stored.

The link to the stored models for the tests can be received from me privately.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: What do you mean? This variable should be renamed to `cached_download`? I have a function called `cached_download`. Can I call this `download_url` instead?

muupan: No, I meant that it should be renamed *because* `cached_download` and `download_and_store_model` use a local variable of the same name.

prabhatnagarajan: Understood.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: Addressed.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: Addressed.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: Addressed. I removed the `CHAINERRL_DOWNLOAD_REPORT`. I had previously walked through the ChainerCV docstring and replaced references to ChainerCV with ChainerRL. The `CHAINERRL_DOWNLOAD_REPORT ` is not relevant to CRL.

muupan: > The `CHAINERRL_DOWNLOAD_REPORT` is not relevant to CRL.

This function's behavior depends on `CHAINERRL_DOWNLOAD_REPORT`, so you should either document it or remove the dependency.

chainerrl/misc/pretrained_models.py
muupan: `Args:` and `Returns:`  should be updated.

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,151 @@
...
+    """Downloads and returns pretrained model.
+
+    Args:
+        alg (string): URL to download from.
+        env (string): Gym Environment name.
+        model_type (string): Either `best` or `final`.
+    Returns:
+        model_path: Path to the downloaded file.
```
muupan:
```suggestion
        str: Path to the downloaded file.
```

chainerrl/misc/pretrained_models.py
```diff
@@ -0,0 +1,151 @@
...
+        is_cached: whether the model was already cached.
```
muupan:
```suggestion
        bool: whether the model was already cached.
```

examples/atari/reproduction/a3c/train_a3c.py
```diff
@@ -137,9 +145,9 @@ def make_env(process_idx, test):
             env=env,
             agent=agent,
             n_steps=None,
-            n_episodes=args.eval_n_runs)
-        print('n_runs: {} mean: {} median: {} stdev: {}'.format(
-            args.eval_n_runs, eval_stats['mean'], eval_stats['median'],
+            n_episodes=args.eval_n_steps)
```
muupan: You meant `n_steps=args.eval_n_steps`?

examples/atari/reproduction/rainbow/README.md
```diff
@@ -21,6 +21,21 @@ python train_rainbow.py [options]
...
+python train_rainbow.py --demo --load-pretrained --env BreakoutNoFrameskip-v4 --pretrained-type best --gpu -1
+```
+
+### Useful Options
+- `--gpu`. Specifies the GPU. If you do not have a GPU on your machine, run the example with the option `--gpu -1`.
+- `--env`. Specifies the environment. 
+- `--demo`. Runs an evaluation, instead of training the agent.
+- `--load-pretrained` Loads the pretrained model. Both `--load` and `--load-pretrained` cannot be used together.
+- `--pretrained-type`. Either `best` (the best intermediate network during training) or `final` (the final network after training).
+
```
muupan: I think `## Running the Example` and `## Running our Pretrained models` should be merged (not only for this example but also for other examples). Having two `### Useful Options`s that largely overlap is not good.

tests/misc_tests/test_pretrained_models.py
```diff
@@ -0,0 +1,564 @@
...
+                           target_update_interval=10 ** 4,
+                           clip_delta=True,
+                           update_interval=4,
+                           batch_accumulator='sum',
+                           phi=lambda x: x)
+
+        model, exists = download_model("DQN", "BreakoutNoFrameskip-v4",
+                                       model_type=self.pretrained_type)
+        agent.load(model)
+        assert exists
```
muupan: Why existence of cache is asserted? It should be false when running this test for the first time.

examples/atari/reproduction/rainbow/README.md
prabhatnagarajan: Ok, will change.

tests/misc_tests/test_pretrained_models.py
prabhatnagarajan: The reason I assert for the existence of the cache is to ensure that we use the cached models during the PFN CI. When running this test locally, indeed it should be false. But when running the test on the CI, we pretrained models should already be downloaded. I omit this test from the Travis CI and only run it in the PFN CI.

muupan: Tests should be able to run locally. Can you remove the assertions?

prabhatnagarajan: Makes sense. I'll remove the assertions. But then we won't have the jobs fail if they caching of trained models isn't done properly, so we should just be vigilant and pay attention to ensure that we're not accidentally downloading models instead of using cached ones.

muupan: That should be ensured definitely, but without breaking the tests locally. I think it is good to have a env var like CHAINERRL_ASSERT_DOWNLOADED_MODEL_IS_CACHED so that we can enable the assertion optionally.

examples/atari/reproduction/a3c/train_a3c.py
prabhatnagarajan: Yep. Fixed.

tests/misc_tests/test_pretrained_models.py
prabhatnagarajan: Done.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: I removed the dependency.

chainerrl/misc/pretrained_models.py
prabhatnagarajan: Done.

PR #569, opened 2019-10-18
Use attrs<19.2.0 with pytest

muupan:
I opened a new PR that is the same as #568 since TravisCI somehow tries to fetch the branch from chainer/chainerrl and fails.


muupan:
Oops, `continuous-integration/travis-ci/push` is still there. Maybe it is checking the commit hash.

PR #570, opened 2019-10-25
Fix weight normalization inside prioritized experience replay

muupan:
This PR fixes the wrong computation of `min_probability` in `PrioritizedReplayBuffer` when `normalize_by_max == 'batch'`, which is the default behavior. The previous behavior was unexpectedly the same as `normalize_by_max == 'memory'`.

### DIFF
--- a/chainerrl/replay_buffers/prioritized.py
+++ b/chainerrl/replay_buffers/prioritized.py
@@ -65,7 +65,7 @@ class PriorityWeightError(object):
     def weights_from_probabilities(self, probabilities, min_probability):
         if self.normalize_by_max == 'batch':
             # discard global min and compute batch min
-            min_probability = np.min(min_probability)
+            min_probability = np.min(probabilities)
         if self.normalize_by_max:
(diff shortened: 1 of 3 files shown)


marioyc:
The test seems to be failing on Python 2

muupan:
Now CI passed.

muupan:
We need to re-evaluate agents, but since the current behavior is clearly wrong, I think it is good to merge it now. By specifying `normalize_by_max == 'memory'` we can retain the current behavior.

ummavi:
Perhaps we could modify existing reproduction scripts with normalize_by_max == 'memory' then?

marioyc:
I guess that would do as a fix, would that need to be part of this PR?

muupan:
> Perhaps we could modify existing reproduction scripts with normalize_by_max == 'memory' then?

That sounds good. I will add that change to this PR.

muupan:
@marioyc Can you check the change I made and merge if it's ok?

PR #581, opened 2019-11-18
Adds proper A3C scores

prabhatnagarajan:
Previously, the A3C results reported the best intermediate score every 1M timesteps. Now it reports  the best intermediate score every 250K timesteps.

### DIFF
--- a/examples/atari/reproduction/a3c/README.md
+++ b/examples/atari/reproduction/a3c/README.md
@@ -21,82 +21,82 @@ python train_a3c.py [options]
 To view the full list of options, either view the code or run the example with the `--help` option.
 
 ## Results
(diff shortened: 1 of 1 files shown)


examples/atari/reproduction/a3c/README.md
```diff
@@ -21,82 +21,82 @@ python train_a3c.py [options]
 To view the full list of options, either view the code or run the example with the `--help` option.
 
 ## Results
-These results reflect ChainerRL  `v0.6.0`. The ChainerRL score currently consists of a single run. The reported results are compared against the scores from the [Noisy Networks Paper](https://arxiv.org/abs/1706.10295), since the original paper does not report scores for the no-op evaluation protocol.
```
muupan: > The ChainerRL score currently consists of a single run.

Why is this sentence removed?

muupan: From offline communication, it is removed because `Number of seeds` is already clarified in the summary below. The removed sentence still holds.

PR #592, opened 2020-02-12
Fix CI errors due to pyglet, zipp, mock, and gym

muupan:
- Use newer gym because old `gym` does not work with pyglet>=1.4.
- Modify some example scripts that are not compatible with newer gym.
- Use zipp==1.0.0 in flexCI because new zipp does not work with Python 3.5.
- Remove mock from dependency as we do not support py2 now.

### DIFF
--- a/.pfnci/run.sh
+++ b/.pfnci/run.sh
@@ -86,7 +86,7 @@ main() {
   # TODO(chainerrl): Prepare test target instead.
   # pytest does not run with attrs==19.2.0 (https://github.com/pytest-dev/pytest/issues/3280)  # NOQA
   "${PYTHON}" -m pip install \
-      'pytest==4.1.1' 'attrs==19.1.0' 'pytest-xdist==1.26.1' 'mock<=3.0.5' \
+      'pytest==4.1.1' 'attrs==19.1.0' 'pytest-xdist==1.26.1' \
       'atari_py==0.1.1' 'opencv-python' 'zipp==1.0.0'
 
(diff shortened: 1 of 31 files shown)


toslunar:
https://travis-ci.org/chainer/chainerrl/jobs/649363317#L2503
```
$ ./test_examples.sh -1
Running example tests: examples_tests/atlas/test_soft_actor_critic.sh
Output files are saved in /tmp/tmp.cdt7eztf9w/atlas/soft_actor_critic/20200212T100236.429911
/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/gym/logger.py:30: UserWarning: WARN: Box bound precision lowered by casting to float32
  warnings.warn(colorize('%s: %s'%('WARN', msg % args), 'yellow'))
  File "examples/atlas/train_soft_actor_critic_atlas.py", line 243, in <module>
    main()
  File "examples/atlas/train_soft_actor_critic_atlas.py", line 123, in main
    timestep_limit = sample_env.spec.tags.get(
AttributeError: 'EnvSpec' object has no attribute 'tags'
The command "./test_examples.sh -1" exited with 1.
```
due to the version of `gym`?

muupan:
@toslunar I added changes to fix errors caused by `gym==0.16.0`. Now CI is all ok. Can you check the additional changes and merge if it's ok?

toslunar:
I confirmed `env.spec.max_episode_steps` is available with gym 0.9.7, the minimum requirement by ChainerRL.

